PR #7, opened 2014-01-07
Allow binding to multiple routing keys.

SebastianEdwards:
Simple patch to allow an array of routing keys to be passed into the queue settings. This allows workers to respond to  selectively respond to multiple topics on a topic exchange. 

E.g.

``` ruby
class Processor
  include Sneakers::Worker
  from_queue :log_processor,
             exchange: 'topic_logs',
             exchange_type: 'topic',
             routing_key: ['kern.*', '*.critical']

  def work(msg)
    err = JSON.parse(msg)
    if err["type"] == "error"
      $redis.incr "processor:#{err["error"]}"
    end

    ack!
  end
end
```

### DIFF
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -29,9 +29,13 @@ class Sneakers::Queue
     handler = @handler_klass.new(@channel)
 
     routing_key = @opts[:routing_key] || @name
+    routing_keys = routing_key.respond_to?(:each) ? routing_key : [routing_key]
 
     queue = @channel.queue(@name, :durable => @opts[:durable])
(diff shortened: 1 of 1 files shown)


selasiehanson:
Has this been tested? Doesn't seem to work.
The routing_key for the exchange still comes out as 'synced_tickets' which is the same as the name of the queue

```ruby
class TicketConsumer 

  include Sneakers::Worker
  from_queue :synced_tickets,   
        exchange: 'app_tickets',
        exchange_type: 'topic',
        routing_key: ['kern.*', '*.critical']        
 
  def work(msg)
    params = JSON.parse(msg)
    p params
    # do processing here
    manual_ack!
  end
end
```

PR #30, opened 2014-03-02
Update bunny version

rodrigosaito:


rud:
Give the gem dependency is specified in the `sneakers.gemfile`, I think this can be safely closed.

rud:
Bunny is at `~> 1.7.0` these days, I believe this can be safely closed @jondot

jondot:
Indeed, thanks @rud !

PR #95, opened 2015-02-09
Making Travis CI badge clickable

rud:
With this it should be possible to click the current build state and jump to the right place on travis-ci.org. One hopes


jondot:
Thanks! :+1:

PR #101, opened 2015-03-12
Redacting the password out of the logging

ngrennan-inflection:


rud:
@ngrennan-inflection, would it be possible to include a few test-cases for this? Also, it seems the `redacted_config` is not actually what is used when logging, so as the patch stands it probably doesn't have an effect? Oh, and `redacted_config = config` makes a reference not a copy of the hash, so you are in effect changing the `config` hash and overwriting the password value with `<redacted>`. A `config.dup` should fix this issue, but the lack of tests will take a bit more.

Hope this is helpful!

rud:
I think this is fixed in the current `master`:
https://github.com/jondot/sneakers/blob/0cb0e08cf395d69358403ba17b10767bfa57ecb5/lib/sneakers/configuration.rb#L93-L94

michaelklishin:
Closing per @rud's comment above.

Issue #105, opened 2015-03-17
Use sneakers with default exchange


abezzub:
Is Sneakers supposed to work with default exchange? If I setup Sneakers like this:
```
Sneakers.configure(amqp: ENV["CLOUDAMQP_URL"],
                   vhost: ENV["CLOUDAMQP_VHOST"],
# other settings...
                   exchange: "")
```
I get this error (partial stack trace):
```
Unexpected error ACCESS_REFUSED - operation not permitted on the default exchange
  /.rvm/gems/ruby-2.2.1/gems/bunny-1.7.0/lib/bunny/channel.rb:1915:in `raise_if_continuation_resulted_in_a_channel_error!'
/.rvm/gems/ruby-2.2.1/gems/bunny-1.7.0/lib/bunny/channel.rb:1073:in `queue_bind'
/.rvm/gems/ruby-2.2.1/gems/bunny-1.7.0/lib/bunny/queue.rb:103:in `bind'
/.rvm/gems/ruby-2.2.1/gems/sneakers-1.0.2/lib/sneakers/queue.rb:39:in `block in subscribe'
/.rvm/gems/ruby-2.2.1/gems/sneakers-1.0.2/lib/sneakers/queue.rb:38:in `each'
/.rvm/gems/ruby-2.2.1/gems/sneakers-1.0.2/lib/sneakers/queue.rb:38:in `subscribe'
/.rvm/gems/ruby-2.2.1/gems/sneakers-1.0.2/lib/sneakers/worker.rb:102:in `run'
```
Error happens here (in queue.rb):
```
    routing_keys.each do |key|
      queue.bind(@exchange, :routing_key => key) # error!
    end
```

If I understand correctly all queues are binded to default exchange by default, so bind can be skipped if `@exchange` is a default exchange. If it is OK to support default exchange I will be happy to submit a PR. I assume something like this should do the trick:
```
    if @opts[:exchange].length > 0
      routing_keys.each do |key|
        queue.bind(@exchange, :routing_key => key)
      end
    end
```
I tested above with my app and it works. Let me know what you think.

michaelklishin:
Default exchange has implicit bindings and is not supposed to be bound explicitly. Sneakers does not take this into account.

sheerun:
A workaround is to set `:routing_key => []` in Sneakers.configure

sheerun:
@abezzub Why did you close?

sheerun:
Oh. I see the commits now.

PR #106, opened 2015-03-17
Allow Sneakers to use default exchange

abezzub:
This PR fixes following issue:

RabbitMQ does not allow to bind queues to default exchange. But Sneakers always tries to bind queue to an exchange. This means that Sneakers can't be used with default exchange. See https://github.com/jondot/sneakers/issues/105

### DIFF
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -22,7 +22,8 @@ class Sneakers::Queue
     @channel = @bunny.create_channel
     @channel.prefetch(@opts[:prefetch])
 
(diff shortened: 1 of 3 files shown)


abezzub:
Second commit fixes build error - "guard" is required for tests but is not specified as dependency

Let me know if you want me to submit second commit in a separate PR

jondot:
Thanks Andrew, I appreciate the guard fix as well! :)
With that in mind - I'd like to start sending Sneakers t-shirts to committers (it is now being designed). Feel free to send me a shipping address to jondotan@gmail.com if you want one :)

PR #107, opened 2015-03-18
Cast exception class to String in maxretry handler

sheerun:
Because for now it always reports `{}`


sheerun:
It also prevents "stack level too deep" error and unacknowledged messages when maxretry worker calls `to_json` on error payload

Issue #89, opened 2015-01-21
Sneakers seems not to respect the vhost parameter from the amqp_url


niwo:
Hi,

I had an issue where the vhost parameter from the amqp_url was not respected.

As example for the amqp_url "amqp://user:password@rabbitmq-host:5672/%2fvhost"
Sneakers would set the vhost to the default value of "/" instead of "/vhost".

Only setting the vhost option explicitly to "/vhost" solved the issue.

Thanks,
Nik

niwo:
As additional information: 

Using Bunny directly works just fine.
Bunny.new with the amqp_url from above takes the correct vhost value: 
```ruby
amqp_url = "amqp://user:password@rabbitmq-host:5672/%2fvhost"
puts Bunny.new(amqp_url)
#<Bunny::Session:9793120 user@rabbitmq-host:5672, vhost=/vhost, hosts=[rabbitmq-host]>
```
I would except the same behavior from Sneakers.

polmiro:
Duplicate from #43. Just manually parse the url and set the vhost. Make sure you have an initial slash on the vhost name.

arrtchiu:
I believe this is a different issue to #43. #43 is about sneakers ignoring an explicit `vhost` param. This is about sneakers not parsing the vhost out of the url correctly. Thankfully @niwo already found a workaround :)

The vhost is in the amqp url and should be parsed correctly. Bunny does parse it correctly.

**To reproduce**:  `Sneakers.configure amqp: 'amqp://user:password@host:5672/myvhost'`

**Expected**:  Sneakers connects to vhost `'myvhost'`

**Actual**:  Sneakers connects to vhost `''`

**Workaround**:  Explicitly set the vhost like this: `Sneakers.configure amqp: 'amqp://user:password@host:5672/myvhost', vhost: 'myvhost'`

Will fix and submit PR hopefully soon.

jondot:
@arrtchiu I'm guessing the problem is around here? https://github.com/jondot/sneakers/blob/64e79bda391f9e9ac9e69944c8249e3552acaddf/lib/sneakers/configuration.rb#L39

arrtchiu:
@jondot Yeah, I think that's the issue - we're setting the `:vhost` key in the `clear` method but not overwriting it later in `merge` which gets called during `Sneakers.configure`. Haven't yet had enough time to verify and patch myself though unfortunately.

arrtchiu:
[...]

PR #108, opened 2015-03-20
parse vhost from amqp url specified in Sneakers.configure

abezzub:
This fixes https://github.com/jondot/sneakers/issues/89.

Before this PR vhost from amqp URL would be ignored when using `Sneakers.configure`:
```
# vhost would not be set
Sneakers.configure({ amqp: "amqp://foo:bar@localhost:5672/foobarvhost" })
```

This PR updates `Configuration` to parse vhost if it is part of amqp URL

### DIFF
--- a/lib/sneakers/configuration.rb
+++ b/lib/sneakers/configuration.rb
@@ -4,7 +4,7 @@ module Sneakers
   class Configuration
 
     extend Forwardable
-    def_delegators :@hash, :to_hash, :[], :[]=, :merge!, :==, :fetch, :delete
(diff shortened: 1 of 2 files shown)


abezzub:
I also tested this in my project's staging server and it connects to CloudAMQP without issues

rud:
Looks great

rud:
In particular, I like how the tests cover the various interesting cases.

michaelklishin:
:+1:

jondot:
Thanks!, looks great! :+1: :+1:

Issue #72, opened 2014-10-08
Mysql2::Error: This connection is still waiting for a result, try again once you have the result


anhuiliujun:
Hello, my worker has reported lots of these errors. After having google the errors, i think it might because two threads shared the same connection.
But i have set only 1 worker with 1 thread. So why could it happen? 
Would you please give me a hand?

jondot:
Closing for cleanups :)

PR #113, opened 2015-04-08
Properly close mysql connections on job timeout

cyoungberg:
https://github.com/brianmario/mysql2/issues/532

If the optional exception class is not passed to Timeout, then the mysql connections will not get cleaned up properly.

This should fix some of the "Mysql2::Error: This connection is still waiting for a result" (#72) issues, if they were a result of a job timing out.


jondot:
Thanks! :)

Issue #112, opened 2015-04-07
Changes today in `thread` gem prevent sneakers `work` method from being called


subvertallchris:
Spent the past few hours troubleshooting processing failures within an app using Sneakers on two different systems. No error messages, the `work` methods in my worker were just never getting called. I had a lot of changes to the app so I worked off the assumption that I had broken it. When I exhausted all the changes, I started looking at dependencies.

`ruby-thread` (`thread` gem) was updated four hours ago. Locking my Gemfile to thread's 0.1.5 version fixed it immediately. Not sure if my config has anything to do with it, let me know if I can help troubleshoot at all.

(Updating the subject, "break sneakers" is such a crappy description.)

inventionlabsSydney:
We're also experiencing this issue, in worker.rb on line 47
```ruby
    def do_work(delivery_info, metadata, msg, handler)
      worker_trace "Working off: #{msg}"

      @pool.process do # Worker is halted here
```

inventionlabsSydney:
Further to this locking my gem file to 0.1.5 of thread fixed this as well.
I'll check on thread's issue tracker and see if anyone else is experiencing on other projects.

rud:
Changes in thread gem in 0.1.6 since 0.1.5: https://github.com/meh/ruby-thread/compare/827c7a180792889ad16cb582cfd66d4a9ec45c2b...master

inventionlabsSydney:
Thanks for that @rud 
It seems that the problem exists with using the `@done_mutex.synchronize `.

I wonder what might be going on here, I am entirely unsure.

rud:
I think this is the issue, was just reported: https://github.com/meh/ruby-thread/issues/34

jondot:
Wow awesome findings guys. I've just merged and cutting a new version!
Thanks!

jondot:
`1.0.3` now in Rubygems

subvertallchris:
Killer, thanks for the jumping on this.

inventionlabsSydney:
Thanks @jondot!

Muchos appreciated!

PR #114, opened 2015-04-08
lock serverengine and thread versions in gemspec

subvertallchris:
Seems necessary considering #112. `thread` doesn't seem to use semantic versioning but `serverengine` does, so it seems safe to set it as is.


jondot:
Awesome, thanks!! :+1: :+1:

PR #115, opened 2015-04-09
Bump thread gem to 0.1.7

### LINKED ISSUE #112: Changes today in `thread` gem prevent sneakers `work` method from being called

nownabe:
thread gem v0.1.7 has be released to fix meh/ruby-thread/issues/34 .
This release also resolves #112 .


nownabe:
I'm sorry for my many commits...
This build failed, but builds seems to fail sometimes independently of commits.

rud:
@nownabe sounds like cross-talk between tests again, state being leaked or setup not being done. The CI setup is still a fairly recent thing, so it makes sense there are still some lingering issues. Any help in that regard would be appreciated, I'm sure, if at all possible.

rud:
@nownabe, the unstable tests should be fixed in #116, no worries.

nownabe:
@rud Thank you!! Test passed!

jondot:
Awesome guys :)

PR #116, opened 2015-04-09
Chore/stabilize tests

rud:
A few minor corrections:
* letting Travis CI pick current ruby-versions in each version-series (2.2, 2.1, etc)
* use `.fetch` to highlight where a `nil` was coming from
* only one `setup` block is run at each level of test; this fixes a few unstable tests. Running `SEED=19690 rake` previously failed, all is well now.


jondot:
Fantastic work! :) :+1: :+1:

PR #120, opened 2015-04-20
Fix max retry count for new version of RabbitMQ x-death format

jaronkk:
As of March 26, RabbitMQ consolidated the x-death header to have only one entry per queue.  A new "count" value was added to track the number of times it's been in a queue. 

I've adjusted the failure_count method so it should work for both the old and new formats for the x-death header.

https://github.com/rabbitmq/rabbitmq-server/issues/78
https://github.com/rabbitmq/rabbitmq-server/pull/79

### DIFF
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -171,7 +171,7 @@ module Sneakers
           if x_death_array.count != 1
             x_death_array.count
           else
-            x_death_array.first['count']
+            x_death_array.first['count'] || 1
           end
         end
       end
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -168,10 +168,12 @@ module Sneakers
           x_death_array = headers['x-death'].select do |x_death|
             x_death['queue'] == @worker_queue_name
           end
-          if x_death_array.count != 1
-            x_death_array.count
+          if x_death_array.count > 0 && x_death_array.first['count']
+            # Newer versions of RabbitMQ return headers with a count key
+            x_death_array.inject(0) {|sum, x_death| sum + x_death['count']}
           else
-            x_death_array.first['count'] || 1
+            # Older versions return a separate x-death header for each failure
+            x_death_array.count
           end
         end
       end
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -165,9 +165,14 @@ module Sneakers
         if headers.nil? || headers['x-death'].nil?
           0
         else
-          headers['x-death'].select do |x_death|
+          x_death_array = headers['x-death'].select do |x_death|
             x_death['queue'] == @worker_queue_name
-          end.count
+          end
+          if x_death_array.count != 1
+            x_death_array.count
+          else
+            x_death_array.first['count']
+          end
         end
       end
       private :failure_count
--- a/spec/sneakers/worker_handlers_spec.rb
+++ b/spec/sneakers/worker_handlers_spec.rb
@@ -103,6 +103,31 @@ describe 'Handlers' do
 
   describe 'Maxretry' do
     let(:max_retries) { nil }
+    let(:props_with_x_death_count) {
+      {
+        :headers => {
+          "x-death" => [
+                        {
+                          "count" => 3,
+                          "reason" => "expired",
+                          "queue" => "downloads-retry",
+                          "time" => Time.now,
+                          "exchange" => "RawMail-retry",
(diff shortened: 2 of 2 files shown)


lib/sneakers/handlers/maxretry.rb
```diff
@@ -165,9 +165,14 @@ def failure_count(headers)
...
         else
-          headers['x-death'].select do |x_death|
+          x_death_array = headers['x-death'].select do |x_death|
             x_death['queue'] == @worker_queue_name
-          end.count
+          end
+          if x_death_array.count != 1
+            x_death_array.count
+          else
+            x_death_array.first['count'] || 1
```
justin-yesware: If I understand, for the "retired once case", using an older version of rabbit, there will be no `count` attribute, and the above `x_death_array` count will be 1, so `x_death_array.first['count']` will return nil, and so you'll return 1, does that sound about right?

Also, what happens when `x_death_array` is 0? Won't this raise an exception? Should there be some protection if that's the case (although it might not be something Rabbit would intentionally do).

spec/sneakers/worker_handlers_spec.rb
```diff
@@ -103,6 +103,31 @@ def process(*args,&block)
...
+                        {
+                          "count" => 3,
+                          "reason" => "expired",
+                          "queue" => "downloads-retry",
+                          "time" => Time.now,
+                          "exchange" => "RawMail-retry",
+                          "routing-keys" => ["RawMail"]
+                        },
+                        {
+                          "count" => 3,
```
justin-yesware: Do these two counts need to be the same? It might be more illustrative to have them be separate so your test below is verifying that you're only using the count associated with the queue you're interested in.

justin-yesware:
Thanks for updating this for newer versions of RabbitMQ! I'm not sure the extra protection is a requirement based on how RabbitMQ works and whether or not the x-death headers are ever not there, but might be worthwhile.

michaelklishin:
It's not one per queue but one per `{queue, reason}`.

lib/sneakers/handlers/maxretry.rb
jaronkk: Your interpretation for the "retired once case" is correct.  It might be a little more clear if the logic was refactored to something like this:

```ruby
if x_death_array.count == 1 && x_death_array.first['count']
  x_death_array.first['count']
else
  x_death_array.count
end
```

For your second question, the x-death headers are always an array.  If there isn't an entry, then it will return x_death_array.count which will be 0, which is the same thing that happens in the current version of the code.  This method is returning the failure count, which will always be 0 the first time it's in the queue.

spec/sneakers/worker_handlers_spec.rb
jaronkk: They wouldn't have to be the same.  In practice the x-death count for both queues is always the same when it is returned to the worker queue, which is why I made them the same here.  I could change the values if you'd prefer.

lib/sneakers/handlers/maxretry.rb
justin-yesware: Yep, gotcha. Sorry I missed that earlier.

spec/sneakers/worker_handlers_spec.rb
justin-yesware: Yeah, that makes sense. I think it's better to keep the test data more realistic, so what you have here is great. Thanks for the explanation. It's been awhile since I've looked into this stuff.

jaronkk:
@justin-yesware The x-death headers aren't there the first time an entry is in the queue, in which case the failure_count method will return 0 (handled by the first if statement in failure_count). 

@michaelklishin That makes sense. 

Would it make sense to change the logic to do a check for a count key, and then sum the counts?

lib/sneakers/handlers/maxretry.rb
```diff
@@ -165,9 +165,16 @@ def failure_count(headers)
...
             x_death['queue'] == @worker_queue_name
-          end.count
+          end
+          if x_death_array.count > 0 && x_death_array.first['count']
+            # Newer versions of RabbitMQ return headers with a count key
+            x_death_array.inject(0) {|sum, x_death| sum + x_death['count']}
+          else
+            # Older versions return a separate x-death header for each failure
+            x_death_array.count
+          end
```
jaronkk: @justin-yesware I updated the logic here so it's more clear when it's using the new vs legacy code.  I also updated it to do a sum of the counts in case there were failures for multiple reasons.

justin-yesware: Changes look good and the internal sum should address the issues @michaelklishin brought up.

Also, this does read a lot cleaner, thanks.

jondot:
Awesome insights guys, thanks @jaronkk!

PR #122, opened 2015-04-24
fix eager loading issue

alucardpj:


cgthornt:
Since Rails isn't a dependency, wouldn't this cause issues for people not using rails?

rud:
The `:environment` pre-requisite rake-task also comes from the Rails part of the world. Good point about the dependency, though.

jondot:
my thoughts are - until we do a Railtie - this will work for people using the tasks on rails, and will explode for people trying to include the tasks via anything other than rails (we also can add some kind of detection of Rails right before that line), but it does not enforce people to install rails along with Sneakers.
What do you think?

tsagadar:
Idea: the call to eager_load! could be guarded by `if defined?(Rails)` to ensure that it only gets executed if Rails is present.

I think this pull request would help a lot of sneakers users - just looking at how many already forked the project to add this line.

jondot:
@tsagadar I completely agree with that idea. I'll revise the PR and merge it in soon.
Thanks!

PR #124, opened 2015-04-30
allow externally-provided Bunny object

dbyron0:
It feels like it oughta be harder than this, but I think this covers it for me.  I'm now calling sneakers like this:
```
    @bunny = Bunny.new('amqp:my_info', logger: Logger.log, heartbeat: Sneakers::Configuration::DEFAULTS[:heartbeat]))
    Sneakers.configure(bunny: @bunny, daemonize: false, log: Logger.log)
```
to get the same info in a Bunny object I used to get when sneakers built it for me.

### DIFF
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -18,8 +18,8 @@ class Sneakers::Queue
   def subscribe(worker)
     # If we've already got a bunny object, use it.  This allows people to
     # specify all kinds of options we don't need to know about (e.g. for ssl).
-    @bunny = @opts[:bunny]
-    @bunny ||= Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
+    @bunny = @opts[:connection]
+    @bunny ||= create_bunny_connection
     @bunny.start
 
     @channel = @bunny.create_channel
@@ -63,4 +63,9 @@ class Sneakers::Queue
     @consumer.cancel if @consumer
     @consumer = nil
   end
+
+  def create_bunny_connection
+    Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
+  end
+  private :create_bunny_connection
 end
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -16,7 +16,10 @@ class Sneakers::Queue
   # :ack
   #
   def subscribe(worker)
-    @bunny = Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
+    # If we've already got a bunny object, use it.  This allows people to
+    # specify all kinds of options we don't need to know about (e.g. for ssl).
+    @bunny = @opts[:bunny]
+    @bunny ||= Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
     @bunny.start
 
     @channel = @bunny.create_channel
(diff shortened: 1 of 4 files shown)


rud:
I think the code change looks good. I think the reason this has not yet been merged is simple. To me the test changes are non-obvious to review as they currently stand. 

Would it be possible to achieve the same goal with a smaller diff, I mean, for the previous behaviour where the bunny object is created internally nothing should have changed. So you should probably only be adding new specs, not changing or moving existing. How does that sound to you?

rud:
Btw, I think it's generally a good way to add this kind of collaborator flexibility, allowing for injected dependencies instead of only creating them internally. Makes it easier to observe and control behaviour in tests too.

lib/sneakers/queue.rb
```diff
@@ -16,7 +16,10 @@ def initialize(name, opts)
   # :ack
   #
   def subscribe(worker)
-    @bunny = Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
+    # If we've already got a bunny object, use it.  This allows people to
+    # specify all kinds of options we don't need to know about (e.g. for ssl).
+    @bunny = @opts[:bunny]
+    @bunny ||= Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
```
rud: One suggestion here: maybe create a private `create_bunny_connection` method, and extract the code 
```ruby
def create_bunny_connection
  Bunny.new(
    @opts[:amqp], 
    :vhost => @opts[:vhost], 
    :heartbeat => @opts[:heartbeat], 
    :logger => Sneakers::logger`
  )
end
```

into it. I mean, that way it's event more obvious what's happening

jondot:
Looks extremely useful! let me know when you're OK to merge this, looks good to me so far!

lib/sneakers/queue.rb
```diff
@@ -16,7 +16,10 @@ def initialize(name, opts)
   # :ack
   #
   def subscribe(worker)
-    @bunny = Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
+    # If we've already got a bunny object, use it.  This allows people to
+    # specify all kinds of options we don't need to know about (e.g. for ssl).
+    @bunny = @opts[:bunny]
```
michaelklishin: I'd rename this to `:connection`.

dbyron0:
Working on making a separate method and naming the config key :connection.  Not sure the best way to handle @rud 's comment about making it more obvious what's changed.  Perhaps viewing the diff without whitespace changes (https://github.com/jondot/sneakers/pull/124?w=1) is enough?  If not, I can think of a couple of different ways to split the commits:

1. one commit to make the code change, but leave the tests completely alone, and another commit to add the tests.
2. one commit to make the code change and add tests, but don't change indentation of existing tests, and another commit that only changes whitespace.

I really like having the tests in the same commit as the code change, so if the link to view diffs without whitespace changes isn't sufficient, my preference is for #2, but I'll do whatever the team wants.

dbyron0:
I think I've made the requested changes in https://github.com/jondot/sneakers/commit/ca5d892ba336a198f7164dd2b8c6b40263b7de6e.  I plan to squash that once it's clear how to structure the previous commits.

rud:
You're right, the diff without whitespace is very readable now: https://github.com/jondot/sneakers/pull/124/files?diff=split&w=0 :+1:

jondot:
Awesome, should I merge away?

dbyron0:
If you're happy with the layouts of the commits I say yes please :)

jondot:
Awesome :) :+1: :+1:

dbyron0:
And now my broken record begins again to request that you please release a version of the gem with this in it.  Thanks much.

PR #130, opened 2015-05-20
sneakers:run rake task bug and fix for it.

otobrglez:
If I tried to run `WORKERS=DoThings rake sneakers:run`on latest sneakers version on my local machine on Ruby 2.2.2. I got following error:

```
rake aborted!
NoMethodError: undefined method `present?' for nil:NilClass
/Users/otobrglez/.rvm/gems/ruby-2.2.2@hunter/gems/sneakers-1.0.4/lib/sneakers/tasks.rb:29:in `block (2 levels) in <top (required)>'
Tasks: TOP => sneakers:run
(See full trace by running task with --trace)
```

So I decided to hunt down this bug and fix it. This is the patch. Hope it helps. :)

### DIFF
--- a/lib/sneakers/tasks.rb
+++ b/lib/sneakers/tasks.rb
@@ -26,7 +26,7 @@ Please set the classes of the workers you want to run like so:
 EOF
       exit(1)
     end
-    opts = (ENV['WORKER_COUNT'].present? ? {:workers => ENV['WORKER_COUNT'].to_i} : {})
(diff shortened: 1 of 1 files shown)


jondot:
yikes!! awesome catch!! thanks :+1: :+1:

shhrikatta:
I was struggling to solve this issue for a long time, this helped me a lot. This works in rails console, but I was wondering why does "present?" not work in rake?

rud:
@shhrikatta I think it's because the `.present?` method is defined in `activesupport` which may not be loaded when the rake-task runs. Docs are here: http://api.rubyonrails.org/classes/Object.html#method-i-present-3F

PR #131, opened 2015-05-21
Change logger for supervisor

nownabe:
Hi, I found problems using with ServerEngine supervisor.

Current Sneakers gives logger to ServerEngine.
This behavior was added by https://github.com/jondot/sneakers/commit/f90e810f5cdc30aaf38fe8f4d39afa3339380b81 .
But in this commit, Sneakers give both :log and :logger options to ServerEngine.

Firstly, I fixed to give only :logger option to ServerEngine.

And when ServerEngine is received a Logger instance as :logger option, ServerEngine aborts.
For detail, Please see https://github.com/fluent/serverengine/pull/20 .

Secondly, I make Sneakers use ServerEngine::DaemonLogger as logger.

### DIFF
--- a/lib/sneakers.rb
+++ b/lib/sneakers.rb
@@ -65,7 +65,7 @@ module Sneakers
     if [:info, :debug, :error, :warn].all?{ |meth| CONFIG[:log].respond_to?(meth) }
       @logger = CONFIG[:log]
     else
-      @logger = Logger.new(CONFIG[:log])
+      @logger = ServerEngine::DaemonLogger.new(CONFIG[:log])
       @logger.formatter = Sneakers::Support::ProductionFormatter
     end
   end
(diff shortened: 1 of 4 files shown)


jondot:
@nownabe many thanks for your contribution! :+1: 
What do you think about `ServerEngine::DaemonLogger` ? is it preferred to use?

nownabe:
@jondot Thank you for your review!
`ServerEngine::DaemonLogger` is a subclass of Logger that is added some special or overrided methods for ServerEngine.
For example, [`ServerEngine::DaemonLogger#reopen!`](https://github.com/fluent/serverengine/blob/master/lib/serverengine/daemon_logger.rb#L185) method is used to re-open log file with `USR2` signal.

I think it is better to use `ServerEngine::DaemonLogger` when using ServerEngine.

jondot:
Thanks @nownabe, I am merging.

nownabe:
Thanks @jondot !

In fluent/serverengine#20, still discussing about logger of ServerEngine.
When I get conclusion, I'll feedback to you.

PR #132, opened 2015-05-21
Allow setting logger and provide server configuration convenience block

pierre-pretorius:
Hi @jondot

People that run Rails often want to set configuration of their workers in initializers.  This configuration should only be active for workers processes and not the when running a normal rails server web process.  Consider what sidekiq provides:

```Ruby
# This block will only yield for sidekiq worker processes.
Sidekiq.configure_server do |config|
  # Place standard Rails output produced when running a job in the worker log file.
  Rails.logger = Sidekiq::Logging.logger
  # Do other config such as adjust DB connection pools, etc. etc.
end
```

This pull request adds two features:
1) A way to override the sneakers logger.
2) A configure_server block method that only yields when running a worker.  Implemented in a common way similar to other gems.

Regards
Pierre

### DIFF
--- a/bin/sneakers
+++ b/bin/sneakers
@@ -2,4 +2,5 @@
 require 'sneakers'
 require 'sneakers/cli'
 
+Sneakers.server = true
 Sneakers::CLI.start
--- a/lib/sneakers.rb
+++ b/lib/sneakers.rb
@@ -47,6 +47,10 @@ module Sneakers
     logger.level = loglevel
   end
 
+  def logger=(logger)
+    @logger = logger
+  end
+
   def logger
     @logger
   end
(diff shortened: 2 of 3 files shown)


jondot:
Thanks @pierre-pretorius this makes a lot of sense. Do you think there is any breaking changes for people who will upgrade?

pierre-pretorius:
There is no breaking changes.  We currently patch your gem to do this in our applications.  If you push a new version to rubygems we can use this change in our application in production for a week or two and then make another pull request later to add the documentation just to be extra safe.

jondot:
Ah! excellent. Thanks @pierre-pretorius :+1: :+1: 
I'm happily merging in :)

PR #134, opened 2015-06-02
Make heartbeat user configurable

salimane:
Currently the logging heartbeat is hardcoded. This pr makes it possible to set it through the config file

### DIFF
--- a/examples/benchmark_worker.rb
+++ b/examples/benchmark_worker.rb
@@ -10,7 +10,8 @@ class BenchmarkWorker
              :prefetch => 50,
              :timeout_job_after => 1,
              :exchange => 'dummy',
-             :heartbeat => 5
+             :heartbeat => 5,
+             :amqp_heartbeat => 10
   def work(msg)
     ack!
   end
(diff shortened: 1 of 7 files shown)


rud:
Good idea. As a constructive suggestion, maybe provide a default value in the config. A test would also go a long way towards making this trivial to merge. Thoughts?

salimane:
@rud there is already a default in the config at https://github.com/jondot/sneakers/blob/master/lib/sneakers/configuration.rb#L25 . But you're right, I should be adding a test example.
Thanks

jondot:
@salimane this would be the AMQP heartbeat, the workers heartbeat relate to server_engine. I think it would make sense to separate the two for cases where people put a 0.5s heartbeat on their AMQP connections (I've seen that happen) - what do you think?

salimane:
@jondot you're right. I've added an additional ``amqp_heartbeat`` config default to 10s.
Thanks

michaelklishin:
FTR, heartbeat values lower than 5 seconds are not recommended: you will get false positives on high latency networks. Hutch used heartbeat value of 1 initially, it was bumped significantly after several production user complaints.

10 seconds sounds like a good default to me.

jondot:
@michaelklishin thanks! sounds good to me too

jondot:
@salimane thanks!, merging in!
:+1:

PR #135, opened 2015-06-03
Add arguments parameter to exchange

RmMsr:
In order to use [custom attributes on an exchange](http://www.rubydoc.info/gems/bunny/Bunny%2FChannel%3Aexchange), I added the option `exchange_arguments` to the configuration. Our use case is the [Delayed Message Plugin](https://github.com/rabbitmq/rabbitmq-delayed-message-exchange).

Thanks


jondot:
Thanks! looks good!

PR #138, opened 2015-06-10
Option to use single thread pool

tsagadar:
Currently sneakers creates one thread pool per worker. With many workers, the number of threads grows quickly, and with it potentially also the number of database connections.

To prevent this and to improve load balancing, set the option "share_threads" to true. This keeps the number of threads to the value given by the option "threads", no matter the number of configured workers.


diddeb:
:+1:

jondot:
Looks like a clean solution! 
Regarding failing tests - https://travis-ci.org/jondot/sneakers/jobs/66219290 
The fix should be just to add a default value for the config (since this is a new field), the tests are aggressive on that.

I'll give you the honor of applying the fix :) either case I'll hop on here in a day or so and if you didn't beat me to it, I'll do it :)

Thanks!

arrtchiu:
:+1: I've been away from this issue for a while, but I agree it looks like this will solve the problem. Nice work!

tsagadar:
Fixed the issues with the tests - missed it doing my first pull request ;-)

jondot:
@tsagadar i'm merging in! thanks!!
:+1:

diddeb:
Sorry, quick question: this is not reflected anywhere in the docs yet – right?

PR #147, opened 2015-07-28
use `from_queue` options in `Worker.enqueue`

luxflux:
Given the following worker:

```ruby
class MyWorker
  include Sneakers::Worker
  from_queue 'swag', exchange: 'other'
end
```

`MyWorker.enqueue 'yolo'` would now publish the message to `sneakers` (the default) exchange. The worker will never pick it up.

This patch fixes this behaviour. Calling `MyWorker.enqueue 'yolo'` will publish a message to an exchange called `other`.

### DIFF
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -143,7 +143,7 @@ module Sneakers
       private
 
       def publisher
-        @publisher ||= Sneakers::Publisher.new
+        @publisher ||= Sneakers::Publisher.new(queue_opts)
       end
     end
   end
--- a/spec/sneakers/worker_spec.rb
+++ b/spec/sneakers/worker_spec.rb
@@ -145,9 +145,12 @@ describe Sneakers::Worker do
         msg.must_equal(message)
         opts.must_equal(:to_queue => "defaults")
       end
+    end
 
-      stub(Sneakers::Publisher).new { mock }
-      DefaultsWorker.enqueue(message)
+    it "passes the configuration to the publisher" do
+      opts = DummyWorker.queue_opts
+      mock(Sneakers::Publisher).new(opts) { mock(Object.new).publish(anything, anything) }
+      DummyWorker.enqueue(message)
     end
   end
 


AlexRiedler:
:+1: I agree this should be done; however, the publishing aspect of this is odd in the general use case.

Not sure if the worker's should have their own connections to each exchange or share them among workers; or this should be removed and made into its own concern.

e.g. there is 5 workers that use the same exchange, but different queues.

or

e.g. there is 5 workers that use a set of exchanges, but the same queue.

@jondot Both of these scenarios seem extremely odd to me in sneakers right now, maybe some insight into design decisions? Or maybe you have opinions on the way to move forward?

jondot:
@AlexRiedler you mention a valid concern. I think, for the publisher, the problem @luxflux is solving is of the default. It's removing the element of surprise - which is always a good thing.

As a design point of view, publishing was always a lower concern for Sneakers (since the crux of the work is for consuming messages), so that typically I've published from Go, Node, and Rails apps (as so other Sneakers users).

So all in all I'm in favor of eliminating the element of surprise - what do you think?

AlexRiedler:
@jondot I agree, lets push this through :)

luxflux:
Has been a while... How can I help to have this merged?

michaelklishin:
This change looks good but I really wonder why `:exchange` is suddenly an option on the `queue` DSL method. It should be a separate method in my opinion.

luxflux:
Since I use sneakers, `:exchange` has always been in the `from_queue` method. This pull request does not add this functionality, it just ensures it is also taken into account for publishing.

gabrieljoelc:
Yes, we've always been able to pass any options we can pass to `Sneakers.configure` to `from_queue` include `exchange_options`. I don't see the issue with making a worker's configuration of subscription and publishing consistent. That said, I do think we need to rethink how what exchanges mean at the endpoint-level. We one endpoint really be subscribing or publishing messages to multiple exchanges?

michaelklishin:
@luxflux fair enough. Then looks like a useful change to me. @jondot WDYT?

jondot:
I agree. let's merge this in!
Thanks :+1:  :+1:

luxflux:
Thanks a lot!

PR #148, opened 2015-08-17
Run all workers by default from CLI or rake. Resolves #110

### LINKED ISSUE #110: `sneakers work` Should run all workers in app/workers
Major nitpick, but it'd be nice if the base command:

`sneakers work`

Looked up workers in `app/workers` or `lib/workers` and ran the ones that are Sneakers workers. That way, when I add a worker, I don't have to remember to update that command.

stevenkaras:
This should make it much more convenient to run all workers declared in an app

### DIFF
--- a/lib/sneakers/tasks.rb
+++ b/lib/sneakers/tasks.rb
@@ -13,9 +13,13 @@ namespace :sneakers do
       ::Rails.application.eager_load!
     end
 
-    workers, missing_workers = Sneakers::Utils.parse_workers(ENV['WORKERS'])
+    if ENV["WORKERS"].nil?
+      workers = Sneakers::Workers::Classes
+    else
+      workers, missing_workers = Sneakers::Utils.parse_workers(ENV['WORKERS'])
+    end
 
-    unless missing_workers.empty?
+    unless missing_workers.nil? || missing_workers.empty?
       puts "Missing workers: #{missing_workers.join(', ')}" if missing_workers
       puts "Did you `require` properly?"
       exit(1)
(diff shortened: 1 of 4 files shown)


jondot:
Looks like a good solution - would you mind taking a look at the failures for the build? (you can ignore the Ruby 1.9.3 one, as we'll be moving that one out)

Thanks!

stevenkaras:
Looks like the test runner isn't unloading the worker after each test case. I think I tested against 2.1 locally, but I'll give it a shot with 2.0 and 2.2, see if I can reproduce the issue.

stevenkaras:
I managed to reproduce this, and also in Ruby 2.1 with test seed 37892. I'll need some help figuring out if it's just unloading the workers properly, or if there's something else wrong here.

gabrieljoelc:
We could really use this. Any way we can move this along?

gabrieljoelc:
thanks! Woot! Green check marks!

gabrieljoelc:
@jondot you willing to merge this now?

jondot:
Sure am. Hey @stevenkaras - happy to see we're on the same time zone :)
Beer on me, whenever you'd like :)

lib/sneakers/tasks.rb
```diff
@@ -13,9 +13,13 @@
       ::Rails.application.eager_load!
     end
 
-    workers, missing_workers = Sneakers::Utils.parse_workers(ENV['WORKERS'])
+    if ENV["WORKERS"].nil?
+      workers = Sneakers::Workers::Classes
```
gabrieljoelc: This should be `Sneakers::Worker::Classes` so it's causing things to break :(

stevenkaras: :scream:

PR #149, opened 2015-08-17
Redact password from inspection of the config object. Resolves #28

### LINKED ISSUE #28: Don't include amqp connection options in logs
Currently, [Sneakers::Worker#worker_trace](https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L108) logs the queue options, which can contain sensitive authentication information (username/password).

The connection details should be scrubbed (only display scheme://host:port/path) prior to being displayed, or omitted entirely (not ideal, as it does provide useful information)

stevenkaras:
This is easier than trying to filter it whenever it is logged, although may not be the cleanest solution.

### DIFF
--- a/lib/sneakers/configuration.rb
+++ b/lib/sneakers/configuration.rb
@@ -71,5 +71,16 @@ module Sneakers
       instance.merge! hash
       instance
     end
+
+    def inspect_with_redaction
+      redacted = self.class.new
+      redacted.merge! to_hash
+
+      # redact passwords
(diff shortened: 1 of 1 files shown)


jondot:
Looks clean to me, that's a nice solution! thanks! :+1: :+1: 
The new AMQP gem requires Ruby 2.0+, before merging I will have to do some checks to see how we can interop, or should we downgrade aqmp.

jondot:
I am merging this in. We'll major bump Sneakers version and support Ruby from 2.0 in the future.

Issue #126, opened 2015-05-08
Idea: controller-like workers


cgthornt:
Hey all, I have an idea and would like some thoughts. I could contribute some time if you like the idea.

It would make development a little bit simpler to take some ideas from rails controllers.

### Idea 1 - Move delivery data to instance methods

Currently Sneakers supports the following syntax:

```ruby
class MyWorker
  include Sneakers::Worker
  
  def work(payload)
    # Do stuff
  end
  
  def work_with_params(payload, delivery_info, metadata)
    # Do stuff
  end
end

```

I think it would make sense to move these to instance methods:

```ruby
class MyWorker
  include Sneakers::Worker
  
  
  def work
    puts "Payload: #{payload}"
    puts "Delivery Info: #{delivery_info}"
    puts "Metadata: #{metadata}"
  end

```

This leads into my next idea.

### Idea 2 - Params

Often content types like JSON and Msgpack are sent over RabbitMQ. Since AMQP supports a 'Content-Type'
header, it might make sense to have the ability to automatically parse them. For example:

```ruby
# Assuming above example is implemented above. Demonstrative:
module Sneakers::Params
  def params
    @params ||= parse_params
  end
  
  # You would want to use some registration pattern below
  def parse_params
    params = nil
    if metadata[:content_type] == 'application/json'
      params = JSON.parse(payload)
    elsif ....
    end
    
    # You could also use strong parameters, hash with indifferent access, etc
    if params && params.is_a?(Hash)
      params = StrongParameters.new(params)
    end
    
  rescue => e
    warn 'Unparseable params' # todo: more robust error handling
  end
  
end

```

You would probably want to make some mime type registration pattern (ommited in code sample above):

```ruby
Sneakers::MIME.register :json, 'application/json'
Sneakers::MIME.register :msgpack, 'aplication/x-message-pack'
Sneakers::MIME.parse :json do |payload|
  JSON.parse(payload)
end
Sneakers::MIME.parse :msgpack, do |payload|
  MsgPack.unpack(payload)
end
```

Now your worker is pretty simple:

```ruby
class MyWorker
  include Sneakers::Worker
  include Sneakers::Params # Maybe include as part of worker?
  
  def work
    puts "Person's email: #{params[:email]}"
    if params[:age] < 21
      puts "Person must be older than 21!"
    end
  end
end
```

Finally, it could be possible to share mime registrations (see idea from #125):

```ruby
Sneakers.publish(:queue_name, 'routing.key', params, format: :json)

[...]

PR #151, opened 2015-08-19
Add a thingy to deserialize JSON

errm:
I was having a look at https://github.com/jondot/sneakers/issues/126 and came up with this as an idea for a patten to implement deserializers. I am just putting this forward really as a discussion point rather than something I expect to be merged in straight away.

* Is this something that should be added like this with an `include` in each worker?  Or would it perhaps be better to be switched on with an option in the config?
* The meta-programming for this is horrible, and it would be much better to do a little refactoring in the worker to expose some hooks for this.


errm:
Infact ignore this, I am going to try and do the MIME registration style interface mentioned in https://github.com/jondot/sneakers/issues/126 I think that would be much cleaner

PR #159, opened 2015-09-03
Fixes Sneakers::Utils#parse_workers for scoped classes. (Fixes #158)

### LINKED ISSUE #158: Scope operator prefixed classes crashes Sneakers::Utils#parse_workers
When attempting to pass a class namespaced under worker I got  an error due to the worker namespace lookup. This is a normal behaviour brought by ruby's way of dealing with namespace lookup. 
The solution then should be to use the scope operator (`::`) for the 'desambiguation' of the class.
When passing `::Worker::Class`, `Sneakers::Utils#parse_workers` tries to find the referred class with this:
```ruby
# ...
w = k.split('::').inject(Kernel){|s, c| s.const_get(c)}
```
which will try to get a const with empty string as its name - which will obviously break.

The solution is simply to use the `Kernel#const_get` directly, unless the split is being used for some corner case.
`w = Kernel.const_get(c)`

AfonsoTsukamoto:


jondot:
Your solution looks like a better solution, can you include a test example where the existing implementation failed for you? i would like to document that

Thanks!

AfonsoTsukamoto:
Ofc. Will do it as soon as I can :+1:

jondot:
Btw this pattern may be common, i just found this as well https://github.com/cinchrb/cinch/blob/master/lib/cinch/utilities/kernel.rb

AfonsoTsukamoto:
I can check those too. The method does work unless the class is scoped.

jondot:
@AfonsoTsukamoto I think I understand the difference Ruby 2.0 makes your solution better, the solution was made to support 1.9. But since this series of Sneakers no longer supports 1.9 -- you are completely right with locating this change!
Happily accepting this, thanks! :)

AfonsoTsukamoto:
Awesome! Still, think there should be a test so the bug doesn't come back. Hope I get some time soon for it :+1:

PR #160, opened 2015-09-03
Fix "log writing failed" error for umlauts

groe:
When the message contains an umlaut, such as "ä", the logging fails with this error:
`log writing failed. "\xC3" from ASCII-8BIT to UTF-8`

Using `String#inspect` the string gets the proper encoding to write it to the log.


rud:
Tiny fix, looks good :+1;

rud:
Meant: :+1:

jondot:
Great!
Thanks :) :+1:

PR #162, opened 2015-09-16
Reformat the readme file

xiangzhuyuan:
when I try to do a demo with sneakers first time, I just confused by readme, so I reformat this file

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,11 @@ gem 'json'
 gem 'redis'
 ```
 
-And a worker
+How do we add a worker? Firstly create a file and name it as `boot.rb`
+then create a worker named as `Processor`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,11 @@ gem 'json'
 gem 'redis'
 ```
 
-And a worker
+How we and a worker? Firstly create a file and name it as `boot.rb`
```
rud: Suggestion for a tweak: first sentence should probably be something like "How do we add a worker?"

xiangzhuyuan: you are right! tks!

jondot:
Thanks! that's very thoughtful towards others!

PR #163, opened 2015-09-23
Removing unmaintained changelog

rud:


jondot:
Yup, that makes sense :+1: :+1:

jondot:
Thanks!

Issue #166, opened 2015-09-29
Specifying Bunny Connection Object Breaks 'sneakers:run'


jcsmith1859:
I have a requirement to use a **TLS** connection to an *AMQP* broker, that also disallows an *auth_mechanism* of type *PLAIN*. I noticed that Sneakers allows for a pre-instantiated *Bunny* object -- which allows me to setup the object in such a way that I can successfully connect to my *AMQP* broker.

The problem that I run into, however, when providing a *connection* option to Sneakers is that there is a piece of code in  [configuration.rb](https://github.com/jondot/sneakers/blob/master/lib/sneakers/configuration.rb#L80) that creates a stack trace before the workers can be setup properly.

The error I receive is:
```Shell
web_1      | rake aborted!
web_1      | NoMethodError: undefined method `sub' for nil:NilClass
web_1      | /usr/local/bundle/gems/sneakers-1.1.2/lib/sneakers/configuration.rb:80:in `inspect_with_redaction'
web_1      | /usr/local/bundle/gems/sneakers-1.1.2/lib/sneakers/runner.rb:62:in `reload_config!'
web_1      | /usr/local/bundle/gems/sneakers-1.1.2/lib/sneakers/runner.rb:11:in `block in run'
web_1      | /usr/local/bundle/gems/serverengine-1.5.10/lib/serverengine/config_loader.rb:39:in `call'
web_1      | /usr/local/bundle/gems/serverengine-1.5.10/lib/serverengine/config_loader.rb:39:in `reload_config'
web_1      | /usr/local/bundle/gems/serverengine-1.5.10/lib/serverengine/config_loader.rb:32:in `initialize'
web_1      | /usr/local/bundle/gems/serverengine-1.5.10/lib/serverengine/daemon.rb:29:in `initialize'
web_1      | /usr/local/bundle/gems/serverengine-1.5.10/lib/serverengine.rb:51:in `new'
web_1      | /usr/local/bundle/gems/serverengine-1.5.10/lib/serverengine.rb:51:in `create'
web_1      | /usr/local/bundle/gems/sneakers-1.1.2/lib/sneakers/runner.rb:11:in `run'
web_1      | /usr/local/bundle/gems/sneakers-1.1.2/lib/sneakers/tasks.rb:38:in `block (2 levels) in <top (required)>'
web_1      | Tasks: TOP => sneakers:run
web_1      | (See full trace by running task with --trace)
```

[...]

PR #167, opened 2015-09-29
166 reload config fix

jcsmith1859:
this is the fix for the issue referenced [here](https://github.com/jondot/sneakers/issues/166).


jondot:
Thanks! merging!

jondot:
there should be a sneakers 2.2.1 now on rubygems with this

PR #168, opened 2015-10-01
allow Sneakers::Publisher to use externally instantiated Bunny::Session

veetow:
`Sneakers::Publisher` should support using the `:connection` option.


veetow:
Hm, not sure why this failed on the 2.2 integration track.  I'm not able to reproduce.

jondot:
It feels like a glitch on the Travis hosted RabbitMQ. On Travis we're running real integration tests. I'll merge this in and force another build

Thanks!! :+1:

PR #169, opened 2015-10-02
Allow more exchange and queue options

### LINKED ISSUE #165: Set more queue options besides name, durable and arguments
I am wanting to be able to set [`auto-delete`](http://rubybunny.info/articles/getting_started.html#hello_world_example) or [`exclusive`](http://rubybunny.info/articles/getting_started.html#weathr_manytomany_topic_routing_example) on queues but can't because `Queue` [doesn't](https://github.com/jondot/sneakers/blob/master/lib/sneakers/queue.rb#L41) allow us to include more options other than the queue name, durability, and the arguments:
```ruby
queue = @channel.queue(@name, :durable => queue_durable, :arguments => @opts[:arguments])
```

gabrieljoelc:
Closes #165

### DIFF
--- a/lib/sneakers/configuration.rb
+++ b/lib/sneakers/configuration.rb
@@ -1,10 +1,18 @@
 require 'forwardable'
+require 'active_support/core_ext/hash/deep_merge'
 
 module Sneakers
   class Configuration
 
     extend Forwardable
     def_delegators :@hash, :to_hash, :[], :[]=, :==, :fetch, :delete, :has_key?
+
+    EXCHANGE_OPTION_DEFAULTS = {
+      :type               => :direct,
+      :durable            => true,
+      :auto_delete        => false,
+      :arguments => {} # Passed as :arguments to Bunny::Channel#exchange
+    }.freeze
 
     DEFAULTS = {
       # runner
@@ -22,13 +30,11 @@ module Sneakers
       :prefetch           => 10,
       :threads            => 10,
       :share_threads      => false,
-      :durable            => true,
       :ack                => true,
       :heartbeat          => 2,
+      :hooks              => {},
       :exchange           => 'sneakers',
-      :exchange_type      => :direct,
-      :exchange_arguments => {}, # Passed as :arguments to Bunny::Channel#exchange
-      :hooks              => {}
+      :exchange_options   => EXCHANGE_OPTION_DEFAULTS
     }.freeze
 
 
@@ -44,6 +50,9 @@ module Sneakers
 
     def merge!(hash)
       hash = hash.dup
+      hash = map_deprecated_exchange_options_key(hash, :exchange_type, :type)
+      hash = map_deprecated_exchange_options_key(hash, :exchange_arguments, :arguments)
+      hash = map_deprecated_exchange_options_key(hash, :durable, :durable)
 
       # parse vhost from amqp if vhost is not specified explicitly, only
       # if we're not given a connection to use.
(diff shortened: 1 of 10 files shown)


lib/sneakers/configuration.rb
```diff
@@ -1,4 +1,5 @@
 require 'forwardable'
+require 'active_support/core_ext/hash/deep_merge'
```
jondot: I'm uncertain we can transparently support `active_support` - it is not in the dependencies for Sneakers, and it might feel a bit heavy as a dependencies for other users of Sneakers. Perhaps there is a way to extract deep merge or use an alternative one?

jondot:
Hi there! awesome to see that there is a project named LA Gear mixing up Sneakers and Sidekiq, very exciting use case! :) 
Would you care to take a look at the notes inlined on the PR before I can merge this in? let me know if I can help in any way :+1:

lib/sneakers/configuration.rb
gabrieljoelc: Yes, I missed that it wasn't in the gemspec. I will try extracting it.

gabrieljoelc:
Huh, the build failed on that commit. They pass locally. Any ideas from the build output?

jondot:
Yes, we're using a real RabbitMQ instance on Travis for real integration tests, and the Sneakers test suite runs at least on 5 configurations in parallel. My bet was that Travis running these in parallel with the same RabbitMQ host would limit connections or some sort.
So now, I've pushed a singular integration test run, builds should run correctly from now on.

gabrieljoelc:
Is there a way to restart this build to check it?

jondot:
Sure just push another commit, they start automatically

jondot:
Or I'll start one manually, hold on

jondot:
Dang. I didn't realize your PR is tracking a tree where the old Travis config still exists.
I'll merge in this PR and we'll watch it pass on master.
Hang tight :)

jondot:
Build is green now!
Thanks for the pull request, looks like a considerable effort in a great direction :)
I'll be pushing a new version to Rubygems momentarily :+1:

gabrieljoelc:
Yay! Thanks!

gabrieljoelc:
@jondot after bumping sneakers in la_gear, I realized that I didn't map deprecated queue arguments, which makes this merge a breaking change for people doing something like:
```ruby
class MyWorker
  include Sneakers::Worker
  from_queue 'my_queue', arguments: { 'x-priority' => 5 }
end
```

jondot:
Ah, wasnt it covered by this?

```
+      hash = map_deprecated_exchange_options_key(hash, :exchange_type, :type)
+      hash = map_deprecated_exchange_options_key(hash, :exchange_arguments, :arguments)
+      hash = map_deprecated_exchange_options_key(hash, :durable, :durable)
```

gabrieljoelc:
No that covers deprecated exchange options and only the queue durable [option](https://github.com/jondot/sneakers/pull/169/files#diff-ee4ef6a4f02dcd635c5738304fa04d0fR105). I clobbered [argument](https://github.com/jondot/sneakers/blob/98433fd9032f8d8c8712062fd917437468d90180/lib/sneakers/queue.rb#L41) option for queues:
```ruby
queue = @channel.queue(@name, :durable => queue_durable, :arguments => @opts[:arguments])
```

gabrieljoelc:
I'm working on a PR that will fix so you won't have to bump major version

gabrieljoelc:
See #174

jondot:
Thanks! I appreciate the quick fix :)
Will be merging right away of course

PR #174, opened 2015-10-06
Make backwards compatible for queue arguments

gabrieljoelc:


jondot:
:+1: :+1: should be on rubygems now as v2.3.1

PR #178, opened 2015-10-09
CI: allow Travis to use the modern infrastructure

rud:


jondot:
awesome stuff @rud ! :+1:

PR #180, opened 2015-10-09
Update the gemspec to the newest version of `thread`.

jagthedrummer:


jagthedrummer:
Instead of `~> 0.2.2` would it be better to do either `~> 0.2.0` or `> 0.1.7`?

rud:
Depends entirely on the gem and their version policy. There has been some trouble with thread in the past, are new issues fixed in this upgrade?

jagthedrummer:
I just updated the gemspec and ran the tests, which report that they have 92%+ coverage and all passed. Has than not been good enough for this particular gem in the past?

gabrieljoelc:
Did you smoke test spinning up a worker? We ran into a problem with thread
that didn't rear its head until we were trying to acknowledge messages with
a real worker.

jagthedrummer:
@gabrieljoelc and @rud, I have run tests with real workers and have not run into any problems so far. I'm trying to find any old issues about upgrading thread, but the only one I find seems to be saying that specifically `0.1.6` was broken. I'm not finding anything about the `0.2.x` series. Is there anything you could point me to?

Just for reference, here's my setup.

In my `Gemfile` I have:

```ruby
gem 'sneakers', :git => 'https://github.com/jagthedrummer/sneakers', :branch => 'jeremy/thread-upgrade'
```

Which gives me this in `Gemfile.lock`

```ruby
GIT
  remote: https://github.com/jagthedrummer/sneakers
  revision: e86de84fef2ad3e7cc4b283574ab71342a61936a
  branch: jeremy/thread-upgrade
  specs:
    sneakers (2.3.4)
      bunny (~> 2.2.0)
      serverengine (~> 1.5.11)
      thor
      thread (~> 0.2.2)
```

Any my simple workers look like this:

```ruby
class PurchaseChargerWorker
  include Sneakers::Worker
  from_queue "#{Rails.env}.purchase.created.charge",
    :exchange => "#{Rails.env}.purchase.created",
    :exchange_options => { :durable => true, :type => :fanout },
    :durable => true

  def work(msg)
    json = JSON.parse(msg)
    puts "json = #{json}"
    purchase = Purchase.find json["purchase_id"]
    PurchaseCharger.new.charge(purchase)
    ack!
  end
end
```

rud:
@jagthedrummer thank you for the manual verification. Probably something we should cover better by automated tests going forward.

gabrieljoelc:
It was this issue I believe https://github.com/meh/ruby-thread/issues/34.

rud:
So, with that, is this particular fear dispelled, or do we need to add more integration-tests to the sneakers project before this can be merged?

gabrieljoelc:
I think we should add more integration tests.

jondot:
Hey guys, quick note - we do have some real integration tests that run on Jenkins after each push, take a look at this: https://github.com/jondot/sneakers/blob/master/spec/sneakers/integration_spec.rb

Keep in mind that this is real integration testing with a real RabbitMQ instance, so many of the operations need to stabilize (the reason for sprinkling sleeps all around). I bet we can make a DSL that encode all these automatically.
In any case this pattern can be used for more integration tests quite easily

gabrieljoelc:
I'm good with that test.

michaelklishin:
If this is still relevant, please rebase against master. Otherwise let's close this :)

PR #181, opened 2015-10-13
Remove incorrect plural from constant name

gabrieljoelc:
Fixes a typo introduced in 7b4079b251ade2301562917be2f8c4edcae619bc that keeps the process from starting unless you set the `WORKERS` environment variable.


jondot:
Thanks for the fix!

PR #184, opened 2015-10-29
Add queue message to #worker_error for better diagnostics

blasterpal:
I am building a slew of microservices and apps using Sneakers and would like to have the actual Rabbit/do_work message logged in `worker_error` so I can trace back the message to the producer. 

Thanks

### DIFF
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -61,11 +61,11 @@ module Sneakers
           end
         rescue Timeout::Error
           res = :timeout
-          worker_error('timeout')
+          worker_error('timeout',msg)
         rescue => ex
           res = :error
           error = ex
-          worker_error('unexpected error', ex)
+          worker_error('unexpected error', msg, ex)
         end
 
         if @should_ack
@@ -109,8 +109,8 @@ module Sneakers
     end
 
     # Helper to log an error message with an optional exception
-    def worker_error(msg, exception = nil)
-      s = log_msg(msg)
+    def worker_error(error_msg, queue_message, exception = nil)
+      s = log_msg("error:'#{error_msg}' message:'#{queue_message}'")
       if exception
         s += " [Exception error=#{exception.message.inspect} error_class=#{exception.class}"
         s += " backtrace=#{exception.backtrace.take(50).join(',')}" unless exception.backtrace.nil?
--- a/spec/sneakers/worker_spec.rb
+++ b/spec/sneakers/worker_spec.rb
@@ -255,7 +255,7 @@ describe Sneakers::Worker do
       handler = Object.new
       header = Object.new
       mock(handler).error(header, nil, "msg", anything)
-      mock(w.logger).error(/unexpected error \[Exception error="foo" error_class=RuntimeError backtrace=.*/)
+      mock(w.logger).error(/error:'unexpected error' message:'msg' \[Exception error="foo" error_class=RuntimeError backtrace=.*/)
       w.do_work(header, nil, "msg", handler)
     end
 
@@ -375,8 +375,8 @@ describe Sneakers::Worker do
     describe '#worker_error' do
       it 'only logs backtraces if present' do
         w = DummyWorker.new(@queue, TestPool.new)
-        mock(w.logger).error(/cuz \[Exception error="boom!" error_class=RuntimeError\]/)
-        w.worker_error('cuz', RuntimeError.new('boom!'))
+        mock(w.logger).error(/error:'cuz' message:'some json' \[Exception error="boom!" error_class=RuntimeError\]/)
+        w.worker_error('cuz', 'some json', RuntimeError.new('boom!'))
       end
     end
 


lib/sneakers/worker.rb
```diff
@@ -61,11 +61,11 @@ def do_work(delivery_info, metadata, msg, handler)
           end
         rescue Timeout::Error
           res = :timeout
-          worker_error('timeout')
+          worker_error('timeout',msg)
         rescue => ex
           res = :error
           error = ex
-          worker_error('unexpected error', ex)
+          worker_error('unexpected error', msg, ex)
```
gabrieljoelc: Seems like the message for an error should go to `logger.debug` not `logger.error`. I'm thinking messages with sensitive data or it filling up logs if you have a handler that keeps retrying and getting an error.

blasterpal: Perhaps we can add a config option , to Sneakers::Configuration? `error_log_message: <boolean>` and have `false` as default? But for my use case, I need to message to debug. I'm not sure how other people are debugging their apps, but I have many queues that flow through each other on various microservices.

gabrieljoelc: We actually have been meaning to PR sneakers with some way to re-raise exceptions so that services like Sentry and Airbrake will pick them up. I think that would be better than logging every thing.

blasterpal: But would the exception include the message?  If so then the application using Sneakers can handle the exceptions and do what it wishes or use Airbrake,et.al.

gabrieljoelc: Sidekiq handles this well for us. When there's an error in a worker, it raises the error including the context of the exception, which has the worker arguments (Sidekiq retires the job then). We use Sentry to capture unhandled exceptions which detects and redacts any sensitive information in the context. This does _not_ add errors or context to the logs, keeping our logs from filling up or leaking sensitive data.

It hasn't mattered as much for us that sneakers doesn't handle this well since we use [`la_gear`](https://github.com/giftcardzen/la_gear) which passes messages straight to Sidekiq from the `Sneakers::Worker#work method` (see [here](https://github.com/giftcardzen/la_gear/blob/develop/lib/la_gear/worker.rb#L68)). However, we have had rare cases that an error was raised in the `Sneakers::Worker` before the message gets into sidekiq, and we are not notified of these errors because sneakers handlers don't have a way to bubble up errors to Sentry. This is is the reason we want to PR sneakers for bubbling that error somehow.

gabrieljoelc: The `raven-ruby` gem is what Sentry uses to hook into our apps, and it includes a [shim](https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L66-L68) for Sidekiq:
```ruby
raven-ruby/lib/raven/integrations/sidekiq.rb
Sidekiq.configure_server do |config|
  config.error_handlers << Proc.new {|ex,context| Raven.capture_exception(ex, :extra => {:sidekiq => context}) }
end
```
And then Sidekiq has a [method](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/exception_handler.rb#L20) that's invoked in their processing rescues that enumerates the `config.error_handlers` collection and passes the context hash to the interface method:
```ruby
# sidekiq/lib/sidekiq/exception_handler.rb
def handle_exception(ex, ctxHash={})
  Sidekiq.error_handlers.each do |handler|
    begin
      handler.call(ex, ctxHash)
    rescue => ex
      Sidekiq.logger.error "!!! ERROR HANDLER THREW AN ERROR !!!"
      Sidekiq.logger.error ex
      Sidekiq.logger.error ex.backtrace.join("\n") unless ex.backtrace.nil?
    end
  end
end
```
Sneakers should use a similar pattern. [`Sneakers::Worker#worker_error`](https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L112-L120) should should enumerate the low-level error handlers (maybe "reporters" instead to avoid ambiguity with the already existing sneakers handlers) collection instead of logging things:
```ruby
# sneakers/lib/sneakers/worker.rb
def worker_error(msg, exception)
  Sneakers.error_reporters.each do |reporter|
    reporter.call(exception, { message: msg }) # add whatever else could be help for the worker context
  end
end
```
We could include a default reporter that copied sidekiq's default error [handler](https://github.com/mperham/sidekiq/blob/7a727c61cfd3f174d32e3a231b14833b61b897fe/lib/sidekiq/exception_handler.rb#L6-L15) and would log the message like you are wanting:
```
# sneakers/lib/sneakers/exception_reporter.rb
module Sneakers
  module ExceptionReporter
    class Logger
      def call(exception, context = {})
        Sneakers.logger.warn(context) if !context.empty?
        Sneakers.logger.warn "#{exception.class.name}: #{ex.message}"
        Sneakers.logger.warn exception.backtrace.join("\n") unless exception.backtrace.nil?
      end
    end
  end
end
```

blasterpal: I agree on this approach of a unified exception Handler/Reporter than can be replaced via configuration,etc. So I can create/extend a JSONReporter for use in my purposes.

PR #187, opened 2015-11-21
Support for fatal (non-retriable) errors

gregbeech:
This commit is an outline of how support for fatal (non-retriable)
errors could be integrated into sneakers in a backwards-compatible way.
This is useful for handling messages that can never succeed (e.g.
poison messages) which at the moment will be retried in the `Maxretry`
handler rather than going straight to the error queue.

Note that although this is (I believe) a functionally complete
implementation the tests are not complete, as I want to gauge interest
and reaction to this before spending too much time on it. If there is
interest then I'm happy to implement all the necessary tests & docs.

Another option (either as well as, or instead of, having `FatalError`)
would be to designate a set of error types as fatal, e.g. when you're
parsing a JSON input you may get a `JSON::ParserError` which you
are not going to want to repeatedly retry.

### DIFF
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -95,9 +95,12 @@ module Sneakers
         end
       end
 
-
       def error(hdr, props, msg, err)
         handle_retry(hdr, props, msg, err)
+      end
+
+      def fatal(hdr, props, msg, err)
+        handle_failure(hdr, props, msg, err)
       end
 
       def timeout(hdr, props, msg)
@@ -128,31 +131,38 @@ module Sneakers
           @channel.reject(hdr.delivery_tag, false)
           # TODO: metrics
         else
-          # Retried more than the max times
-          # Publish the original message with the routing_key to the error exchange
-          Sneakers.logger.info do
-            "#{log_prefix} msg=failing, retry_count=#{num_attempts}, reason=#{reason}"
-          end
-          data = {
-            error: reason,
-            num_attempts: num_attempts,
-            failed_at: Time.now.iso8601,
-            payload: Base64.encode64(msg.to_s)
-          }.tap do |hash|
-            if reason.is_a?(Exception)
-              hash[:error_class] = reason.class
-              hash[:error_message] = "#{reason}"
-              if reason.backtrace
-                hash[:backtrace] = reason.backtrace.take(10).join(', ')
-              end
-            end
-          end.to_json
-          @error_exchange.publish(data, :routing_key => hdr.routing_key)
-          @channel.acknowledge(hdr.delivery_tag, false)
-          # TODO: metrics
+          handle_failure(hdr, props, msg, reason, num_attempts)
         end
       end
       private :handle_retry
+
+      def handle_failure(hdr, props, msg, reason, num_attempts = nil)
+        # +1 for the current attempt
+        num_attempts ||= failure_count(props[:headers]) + 1
+        # Retried more than the max times
+        # Publish the original message with the routing_key to the error exchange
+        Sneakers.logger.info do
+          "#{log_prefix} msg=failing, retry_count=#{num_attempts}, reason=#{reason}"
+        end
+        data = {
+          error: reason,
+          num_attempts: num_attempts,
+          failed_at: Time.now.iso8601,
+          payload: Base64.encode64(msg.to_s)
+        }.tap do |hash|
+          if reason.is_a?(Exception)
+            hash[:error_class] = reason.class.to_s
+            hash[:error_message] = "#{reason}"
+            if reason.backtrace
+              hash[:backtrace] = reason.backtrace.take(10).join(', ')
+            end
+          end
+        end.to_json
+        @error_exchange.publish(data, :routing_key => hdr.routing_key)
+        @channel.acknowledge(hdr.delivery_tag, false)
+        # TODO: metrics
+      end
+      private :handle_failure
 
       # Uses the x-death header to determine the number of failures this job has
       # seen in the past. This does not count the current failure. So for
--- a/lib/sneakers/handlers/oneshot.rb
+++ b/lib/sneakers/handlers/oneshot.rb
@@ -18,6 +18,10 @@ module Sneakers
         reject(hdr, props, msg)
       end
 
+      def fatal(hdr, props, msg, err)
+        error(hdr, props, msg, err)
+      end
+
       def timeout(hdr, props, msg)
         reject(hdr, props, msg)
       end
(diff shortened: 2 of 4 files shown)


gabrieljoelc:
How do you think could play into #188?

gregbeech:
@gabrieljoelc I think you could potentially go a couple of ways on this.

As that ticket is currently written with error reporters the concerns could be treated as orthogonal, and you could add it alongside with the reporters being called in the worker before the handler.

However, a more interesting approach might be to allow different error handling strategies to be plugged into the handler's `#error` method instead of having a separate `#fatal` handler, e.g. something like the following (but with better naming etc.)

~~~ruby
class Maxretry
  # ...

  def error(hdr, props, msg, err)
    case error_strategies.find { |strategy| strategy.call(msg, err) }
    when :fail then handle_failure(hdr, props, msg, err)
    else handle_retry(hdr, props, msg, err)
    end
  end

  # ...
end

class LoggingStrategy
  def call(msg, err)
    logger.log(msg, err)
    nil # continue the chain
  end
end

class MyFatalStrategy
  def call(msg, err)
    return :fail if err.is_a?(JSON::ParserError) # fail on your app's fatal errors
    nil # otherwise continue the chain
  end
end
~~~

This would need to be more robust (e.g. handling the case where error strategies themselves threw errors) but I've kept this as simple as possible to demonstrate the concept. Handlers would run in order so you'd need to make sure that the logging strategy was before the fatal strategy in the example above.

This would give a neat way to allow each application to define which errors are fatal for it, and would keep the core of sneakers simpler than the approach I've currently proposed, which would be a good thing. The only downside I could see is that you'd _have_ to `raise` an error rather than returning `:fatal` to trigger the behaviour, but that _probably_ isn't the end of the world as a fatal error would typically be on something like a poison message which you'd hope would be a rare occurrence in well designed systems.

gabrieljoelc:
I think error reporters should be enumerated in the worker and not in handlers. _Handlers_ are responsible for handling messaging semantics (kinda like rails controller actions handle http method request semantics). _Error reporters_ are responsible for reporting exceptions (logs, Sentry, Airbrake, etc.).

gregbeech:
Yeah, that's a fair point, I agree that it would be more appropriate to handle any error reporting in the worker rather than the message handler. So in that case you could:

- Either treat #188 as orthogonal and have the error reporters enumerated in the worker (for both normal or fatal errors) and support fatal errors as illustrated in this PR.

- Use the error strategy approach outlined above but enumerate them in the worker, and allow them to decide whether an error is fatal or not (and then the worker would call either `#error` or `#fatal` on the messaging handler as appropriate).

I think either of these approaches is fine. The first is a little simpler, the latter is a little more powerful.

Of course there's the option to not include support for classifying errors as fatal in sneakers. I don't think that's a great option because poison messages _et al_ are a basic fact of life that everyone has to deal with (and are typically a different kind of error to normal transient errors) so it would be nice if the library helped out.

gabrieljoelc:
> Use the error strategy approach outlined above but enumerate them in the worker, and allow them to decide whether an error is fatal or not (and then the worker would call either #error or #fatal on the messaging handler as appropriate).

I still think this approach is mixing the responsibilities of handlers and error reporters because the error reporters are making decisions on the `error` or `fatal`, which are part of sneakers [message semantics](https://github.com/jondot/sneakers/wiki/Message-handling-semantics).

Can you think of other use cases that would use the `fatal` besides poison messages?

gregbeech:
> I still think this approach is mixing the responsibilities of handlers and error reporters because the error reporters are making decisions on the error or fatal, which are part of sneakers message semantics.

Yep that's fair enough, it was really just a suggestion. In that case I guess the simple approach taken in this PR of assuming that any reporting of errors is an orthogonal concern would be the 'right' one.

> Can you think of other use cases that would use the fatal besides poison messages?

Poison messages is really the main one. But when you consider that a poison message is any message for which processing can never succeed, that covers a fairly wide variety of cases, e.g.

- Invalid message structure
- Invalid message data (e.g. nonexistent entity id)
- No longer valid message data (e.g. trying to update an out-of-date entity)

A valid strategy for _some_ of these messages is just to ack them and drop them on the floor (e.g. for no-longer-valid message data then simply logging a warning and dropping the message is quite possibly reasonable) but for others you'd want to know about the error because it could indicate a bug in your application.

The idea of `#fatal` is to give you a better way of reporting a 'real' error in these situations, as at the moment neither of the choices are great:

- Log it from your worker implementation and ack the message, which isn't a great separation of concerns (and makes for weird tests "when given an invalid message it should ack"!). It also somewhat negates the value of #188 if implemented because the reporters would never be called as the message appears to succeed.

- Raise an error and allow the message to go round the retry loop N times, which isn't ideal because it wastes processing time and makes your retry queue larger with messages that won't succeed. It'll also clog up your logs with the repeated errors.

gabrieljoelc:
Ok, that makes sense. So this _would_ be adding a new messaging semantic to sneakers. This [wiki](https://github.com/jondot/sneakers/wiki/Message-handling-semantics) section would need to be updated:
> Sneakers supports these directly, and you can use the matching methods in your worker

> `ack!`
> `reject!`
> `requeue!`
> Internally, sneakers will also report a couple more events

> `error` for unexpected errors
> `noop` when all went well, but no `ack!` was given

where this would go at the end:
> `fatal` expected error occurred (i.e. poison messages)

gregbeech:
Yep there would be a few docs needing to be updated, just wanted to make sure the maintainers were happy with both the concept and approach before I put too much time into writing all the necessary tests and updating all the necessary docs as that's probably a good few hours rather than the short time I spent on this.

There are still a couple of open questions I have around the implementation and naming, specifically:

- Should the 'fatalness' of an error be passed to the `#error` handler method as an additional argument (i.e. `fatal = false`) rather than introducing a new `#fatal` handler method? The reason I didn't do this was for backwards compatibility with existing custom handlers people may have written, but this could also be done based on the arity of the method.

- Is the name `fatal` the right one? My slight concern is that this is the same name as Ruby's inbuilt `fatal` error type for really serious system errors, so perhaps this would be better called something like `:fail` instead.

jondot:
Great discussion :+1:.  Indeed poison messages are important to fend off once identified - I'm interested to learn about what is an actual example of such a message in your infrastructure / actual code?
Regarding fatal/fail - fail sounds more reasonable, as your intuition already spotted.

I think there is no harm adding more semantics to the handlers, as long as they represent a real pain more than one person have bumped into (so that to make sure it is at least somewhat a general pain).
And as @gabrieljoelc states (great analogy there!) handlers are the way you want to wire messaging semantics.

Happy thanksgiving!

michaelklishin:
If this is still relevant, please rebase against master and respond to @jondot's comment above :)

gabrieljoelc:
I think this should be closed because are often relative to the application's interpretation and should be dealt with by a custom handler.

Issue #192, opened 2015-12-02
Sneakers::Publisher leaves channels open while still creating new ones


veetow:
Channels never get closed.  Consider using `Bunny::Session#with_channel` for auto-closed, temporary channels.

http://www.rubydoc.info/github/ruby-amqp/bunny/master/Bunny%2FSession%3Awith_channel

michaelklishin:
Not closed in what context? Can you please elaborate?

veetow:
Sorry, I meant to add details yesterday, but I got busy at work.  It appears to be isolated to (maybe just the way I'm using) `Sneakers::Publisher#publish`.

In `Sneakers.configure`, I am passing in an existing `Bunny::Session`.

In my publisher class I have a class method that publishes some data on a schedule:

```
def self.publish(payload)
  Sneakers::Publisher.new(self.publish_options).publish(payload, self.publish_options)
end
```

Each time the schedule fires, a new channel is opened on the existing connection and the grow without getting closed.

veetow:
If I do this, all is right in the world:

```
def self.publish(payload)
  p = Sneakers::Publisher.new(self.publish_options)

  p.publish(payload, self.publish_options)
  p.instance_variable_get(:@channel).close
end
```

rud:
The Publisher class  is not really useful for anything than a quick demo of how to kind of use Bunny. Every time you create an instance you get a new connection to RabbitMQ - https://github.com/jondot/sneakers/blob/081208977eb5d606b1b7f6f3e24f83410e2c305b/lib/sneakers/publisher.rb#L10

So, a quick fix would be to reuse the instance, or create your own publisher class that can take different options for each time you publish.

rud:
That said, the documentation for the `Publisher` class should probably highlight this, you are not the first to be burned by this code.

veetow:
@rud, I think it can be polished up a bit to work well enough for general case stuff.  I'm working on a pull request now.  Something to the effect of this:

```
diff --git a/lib/sneakers/publisher.rb b/lib/sneakers/publisher.rb
index ec49038..719b9bd 100644
--- a/lib/sneakers/publisher.rb
+++ b/lib/sneakers/publisher.rb
@@ -12,7 +12,10 @@ module Sneakers
       to_queue = options.delete(:to_queue)
       options[:routing_key] ||= to_queue
       Sneakers.logger.info {"publishing <#{msg}> to [#{options[:routing_key]}]"}
-      @exchange.publish(msg, options)
+
[...]

PR #193, opened 2015-12-03
use #with_channel rather than maintaining a persistent channel and exchange

veetow:
https://github.com/jondot/sneakers/issues/192

### DIFF
--- a/lib/sneakers/publisher.rb
+++ b/lib/sneakers/publisher.rb
@@ -12,11 +12,11 @@ module Sneakers
       to_queue = options.delete(:to_queue)
       options[:routing_key] ||= to_queue
       Sneakers.logger.info {"publishing <#{msg}> to [#{options[:routing_key]}]"}
-      @exchange.publish(msg, options)
+
+      @bunny.with_channel do |channel|
+        channel.exchange(@opts[:exchange], @opts[:exchange_options]).publish(msg, options)
+      end
     end
-
-
-    attr_reader :exchange
 
   private
     def ensure_connection!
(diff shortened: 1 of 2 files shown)


rud:
Ow, looks like a lot of the tests didn't behave too well. The change itself looks good to me. Anyone else?

michaelklishin:
Using `with_channel` isn't critically important. I don't really object to it either.

What worries me is how many tests use mocks. Bunny uses no mocks.

veetow:
Previously the `Bunny::Session` instance was only used in `#ensure_connection!` which in most examples was mocked out of the execution path (`mock(p).ensure_connection! {}`).  So now the extra mocking of Bunny is required due to the usage of `@bunny.with_channel` in the `#publish` method.  We're unit testing Sneakers here, not Bunny, so IMO these mocks are required.

jondot:
@veetow question - do you somehow pool/cache your publisher? I'm wondering if you are still happy with a cached publisher that uses a single persistent connection. One that closes/opens one per message sounds a bit strange, I'm not sure what's the effect of this on RabbitMQ...

veetow:
@jondot, I was not keeping a persistent publisher because I assumed that configuring a persisted Bunny::Session was good enough.  I've made some improvements to my code since, but I still think that using `#with_channel` here is more flexible and fool proof.  There's really no advantage to a long-lived channel.

jondot:
Thanks @veetow, I am still wondering if opening/closing the connection for high rate of publishing repeatedly is thrashing RMQ. @michaelklishin is there any such effect that you know of by a chance ?

veetow:
You mean by opening and closing the channel (not the connection)? Nope, as channels are multiplexed across the single connection.  Via `Sneakers.configure` I am passing in an existing `Bunny::Session` so there is no connection tear-down/set-up.

michaelklishin:
This goes against how most messaging protocols are supposed to be used, and AMQP 0-9-1 in particular. I understand what kind of improvement this offers but opening a new channel per message will severely drop throughput (as it is a network roundtrip before you can do anything, in fact, in this case it is two because the new channel instance won't have the exchange cached!).

PR #194, opened 2015-12-17
Maxretry more flexible

nevesenin:
Hi,

we are currently evaluating if we can switch from our own bunny wrapper gem to sneakers. To make sneakers fit to our existing infrastructure the MaxRetryHandler needs some more configuration options. I'd like to discuss the idea.

Merten


jondot:
Hi Merten,
The handlers were designed in a modular way such that you can be sure you can use what ever handlers you'd like with Sneakers.

Generally there are 3 ways people can contribute handlers:

1. Commit to core handlers (in this case, let's just get a thumbs up from one more person using MaxRetry)
2. Add handlers as you see fit to core Sneakers (even if they are variants of existing ones, so you could call this one for example AdvancedMaxRetry)
3. If the handlers you use contain sensitive logic / IP, you can maintain your own repo like a sneakers-handlers or put them into your worker repo in a folder which contains any handler you would like

So what I'm trying to say - any form of a handler contribution is legitimate, so feel free to keep them coming :)

Regarding current pull request - @rud  @justin-yesware what do you think?

rud:
Thank you for the ping @jondot. A few comments:

* The current `MaxRetry` is indeed quite rigid, and these changes do look like they would fix some of that rigidity. I'd say some integration test scenarios would be in order as this is tricky territory to modify, and I totally understand they're not the first order of business. The current code here looks like it pretty much defaults the the current behaviour, is that correctly understood?
* I'm all for modernizing or cleaning up code style, but I'd personally prefer it as a separate pull-request from semantic changes. I'd suggest using `rubocop` for consistency, it's quite neat for this sort of cleanup work.

Hope this is helpful.

nevesenin:
Thanks for the feedback. Yes, the current behavior doesn't change with the defaults. I aggree to not include the style changes and I'll add an integration test scenario. I also like rubocop and I would like to make another pull request for style changes.

lib/sneakers/handlers/maxretry.rb
```diff
@@ -59,23 +66,25 @@ def initialize(channel, queue, opts)
...
+
+        @retry_queue = @channel.queue(retry_queue_name,
+                                      :durable => queue_durable?,
+                                      :arguments => {
                                        :'x-dead-letter-exchange' => requeue_name,
-                                       :'x-message-ttl' => @opts[:retry_timeout] || 60000
-                                     })
-        @retry_queue.bind(@retry_exchange, :routing_key => '#')
+                                       :'x-message-ttl' => @opts[:retry_timeout] || 60000,
+                                       :'x-dead-letter-routing-key' => requeue_routing_key
```
justin-yesware: This will change the current behavior I believe. My understanding of this setting is that if it is unset, the routing-key from the originally published message will be used in the retry queue. When the TTL is hit and it goes to the dead-letter exchange (which will then forward it onto the original queue sneakers is subscribing to), it will again pass along the original message's routing key. By setting the value, you're over-writing the original message's routing key, potentially causing the retried message to be different than when it first went through the queue.

I'm not an expert in this area of Rabbit, as we never don't use routing keys, but I suspect this will cause some configurations to not work. I think a test case of this would be to setup your Sneakers worker to use a queue that binds to a topic exchange with a rule to only get messages that have a routing key of "hello". Then publish a message with that routing key, have your worker fail it so it goes through this retry logic, and set the value for this requeue routing key to be "world" and see if you get it again in your worker. If you do, I think a further validation would be to verify that that message you're getting has the routing key value of "hello" and not "world". My suspicion is that you'll never see the retry and if you do, it will have a routing key value of "world".

justin-yesware:
I like the additional configuration options this pull has, and I like the direction this is headed, but fear it will break cases that rely on routing key-based queue binding. An integration test should surface that so +1 to that.

lib/sneakers/handlers/maxretry.rb
```diff
@@ -59,23 +66,25 @@ def initialize(channel, queue, opts)
...
         end
-        @error_queue = @channel.queue(error_name,
+        @error_queue = @channel.queue(error_queue_name,
                                       :durable => queue_durable?)
-        @error_queue.bind(@error_exchange, :routing_key => '#')
+        @error_queue.bind(@error_exchange, :routing_key => @error_routing_key)
 
         # Finally, bind the worker queue to our requeue exchange
-        queue.bind(@requeue_exchange, :routing_key => '#')
+        queue.bind(@requeue_exchange, :routing_key => requeue_routing_key)
```
nevesenin: Here you get the messages with "world" routing key.

nevesenin:
I'll add a test as soon as possible.

nevesenin:
Added integration tests.

spec/fixtures/maxretry_worker.rb
```diff
@@ -0,0 +1,45 @@
+require 'sneakers'
+require 'thread'
+require 'redis'
+
+require 'sneakers/handlers/maxretry'
+
+# This worker ... works
```
justin-yesware: I think I would call these workers something more along the lines of how they behave: `AlwaysAckWorker`, `AlwaysRejectWorker`, `RejectOnceWorker` (which is what you have).  It may make them more portable across other tests.

spec/fixtures/maxretry_worker.rb
```diff
@@ -0,0 +1,45 @@
...
+  def work_with_params(_, delivery_info, message_properties)
+    if message_properties[:headers].nil? ||
+       message_properties[:headers]['x-death'].nil?
+      reject!
+    else
+      dump = JSON.dump(
+        'delivery_info' => delivery_info.to_hash,
+        'message_properties' => message_properties.to_hash
+      )
+      Redis.new.set(
```
justin-yesware: Any reason not to use a Redis list here? Could be useful to test multiple payloads.

spec/sneakers/integration_spec.rb
```diff
@@ -100,8 +85,352 @@ def start_worker(w)
...
+          refute_nil(message.first)
+        end
+
+        it 'has been retried twice' do
+          # wait for failing message
+          sleep REQUEUING_WAIT_TIME
+
+          message = get_message_from_queue("#{queue_name}-error")
+
+          # Hmmmmm... should be 2
```
justin-yesware: The 3 is correct here. It is retried twice, which means a worker sees it 3 times total, yielding the 3 attempts.

spec/sneakers/integration_spec.rb
```diff
@@ -100,8 +85,352 @@ def start_worker(w)
...
+        it 'has no message in the error queue' do
+          # wait for failing message
+          sleep REQUEUING_WAIT_TIME
+
+          message = get_message_from_queue("#{queue_name}-error")
+
+          assert_nil(message.first)
+        end
+
+        it 'consumes the requeued message' do
```
justin-yesware: I think it would be good to ensure in this test that the retried message has the same routing key that it originally did. My initial concern was that by always providing a routing key on the dead letter exchange, the original routing key will be replaced when it is retried. I think for the case where you want to override each of these values, that is ok, but for the nominal case, you want to preserve the original message as much as possible.

spec/fixtures/maxretry_worker.rb
nevesenin: Good point. Done.

spec/fixtures/maxretry_worker.rb
nevesenin: I don't need a list for that.;)

spec/sneakers/integration_spec.rb
nevesenin: Uhm. I think I mixed up attempts and retries.

spec/sneakers/integration_spec.rb
nevesenin: Ok, now I got your point. I never thought about a scenario where routing keys are important for the consumer. I see routing keys as information for rabbitmq. The worker needs to know the routing key, only because it creates the binding between exchange and queue. It consumes every message in the queue no matter how it gets there.
The retried message does not have the same routing key as the original message. There is an additional binding which routes the retry messages to the original queue.
I added some integration tests to show how the maxretry handler would work in our existing environment. It creates only one additional exchange. The error exchange. This is why we need different routing keys.
If it is important to keep the original routing key, I think I should add a new
handler instead of changing this one.

spec/fixtures/maxretry_worker.rb
nevesenin: I mean, if someone needs a list, it can be changed easily. If I change it now, one could wonder why I use a list when a string is enough.

adisos:
Sorry for dummy question, but what should I do with messages in dead queue? I used to sidekiq before, so there it's easy to requeue failed jobs.

nevesenin:
I close this pull request in favor of https://github.com/jondot/sneakers/pull/205.

PR #201, opened 2016-01-29
Adding attr_reader of channel into queue class so that it maybe acces…

inventionlabsSydney:
This pull request it to add the channel to attr_reader on the Sneakers::Queue class.

We have a particular use case that we deem would also be useful for other people which is accessing the default exchange for publishing within a task.
In our scenario we have a task system that publishes a task to a routed topic, this packet also includes a reply_to queue and correlation_id which when the task is finished in our worker reports the result to that.

To give you an example of this worker framework, I have attached what we use:

```
require_relative 'framework_helper'
module AntHQ
  module WorkerHelper
    def work_with_params(msg, delivery_info, metadata)
      @task_success = true

      raise ArgumentError, "Task method [perform_work(msg, delivery_info, metadata)] undefined, this must be declared" unless defined? perform_work
      # 
      # Do the work in the task
      # 
      response = perform_work(msg, delivery_info, metadata)
      #
      # Ensure that the response is acceptable
      # 
      unless response.kind_of? ResponseObject::ResponseObjectSuccess or response.kind_of? ResponseObject::ResponseObjectFailure
        raise PerformWork::Error::DidNotReturnResponseObject, "The perform_work method did not return a ResponseObject" 
      end
    rescue Exception => the_error
      #
      # Whenever an exception is raised, lets always ensure it's handed to Rollbar
      #
      Rollbar.error(the_error, error_pickup: "Core Daemon Exception", payload_message: msg, delivery_info: delivery_info, worker_metadata: metadata)
      @task_success = false
      #
      # Ensure we build a response object to pass to reply_to if it's applicable
      # 
      response = ResponseObjectFailure.new(the_error)
    ensure
      #
      # If metadata contains the key :reply_to then we deliver the response object to the reply_queue via the default exchange
      #
      if metadata.reply_to and metadata.correlation_id
        #
        # Sneakers::Queue class was modified to allow access to channel
        #
        @queue.channel.default_exchange.publish(response.to_reply, routing_key: metadata.reply_to, correlation_id: metadata.correlation_id)
      end
      #
      # Task was successful, lets ack and move on.
      # 
      return ack! if @task_success
      #
      # If this packet is rejected we will pass to the Dead Letter Exchange for evacuation and ensurance.
      # 
      return reject! unless @task_success
    end
  end
end
```
This framework is included into a worker like so:

```

module AntHQ
  module Worker
    class Product
      include Sneakers::Worker
      include AntHQ::WorkerHelper
      from_queue :testing_tasks

      def perform_work(msg, delivery_info, metadata)
        puts "I'm doing some task work in here yay!"
        return ResponseObject.success({test: true})
      end
    end
  end
end
```

The particular line this affects is: `@queue.channel.default_exchange.publish(response.to_reply, routing_key: metadata.reply_to, correlation_id: metadata.correlation_id)`

Which without this change can only access attr_reader :exchange which may be bind in such a way that this packet could not be delivered to the correct reply_to queue.

I hope this is a reasonable explanation!

By all means if there's a better way to do this, please let me know :+1: 

Thanks,
--Karl.


inventionlabsSydney:
I am closing this pull request due to not having an up to date fork. I will recommit this pull request.

PR #203, opened 2016-02-16
Adding attr_reader of channel into queue class so that it maybe accessible via queue inside worker

inventionlabsSydney:
This pull request it to add the channel to attr_reader on the Sneakers::Queue class.

We have a particular use case that we deem would also be useful for other people which is accessing the default exchange for publishing within a task.
In our scenario we have a task system that publishes a task to a routed topic, this packet also includes a reply_to queue and correlation_id which when the task is finished in our worker reports the result to that.

To give you an example of this worker framework, I have attached what we use:

```
require_relative 'framework_helper'
module AntHQ
  module WorkerHelper
    def work_with_params(msg, delivery_info, metadata)
      @task_success = true

      raise ArgumentError, "Task method [perform_work(msg, delivery_info, metadata)] undefined, this must be declared" unless defined? perform_work
      # 
      # Do the work in the task
      # 
      response = perform_work(msg, delivery_info, metadata)
      #
      # Ensure that the response is acceptable
      # 
      unless response.kind_of? ResponseObject::ResponseObjectSuccess or response.kind_of? ResponseObject::ResponseObjectFailure
        raise PerformWork::Error::DidNotReturnResponseObject, "The perform_work method did not return a ResponseObject" 
      end
    rescue Exception => the_error
      #
      # Whenever an exception is raised, lets always ensure it's handed to Rollbar
      #
      Rollbar.error(the_error, error_pickup: "Core Daemon Exception", payload_message: msg, delivery_info: delivery_info, worker_metadata: metadata)
      @task_success = false
      #
      # Ensure we build a response object to pass to reply_to if it's applicable
      # 
      response = ResponseObjectFailure.new(the_error)
    ensure
      #
      # If metadata contains the key :reply_to then we deliver the response object to the reply_queue via the default exchange
      #
      if metadata.reply_to and metadata.correlation_id
        #
        # Sneakers::Queue class was modified to allow access to channel
        #
        @queue.channel.default_exchange.publish(response.to_reply, routing_key: metadata.reply_to, correlation_id: metadata.correlation_id)
      end
      #
      # Task was successful, lets ack and move on.
      # 
      return ack! if @task_success
      #
      # If this packet is rejected we will pass to the Dead Letter Exchange for evacuation and ensurance.
      # 
      return reject! unless @task_success
    end
  end
end
```
This framework is included into a worker like so:

```

module AntHQ
  module Worker
    class Product
      include Sneakers::Worker
      include AntHQ::WorkerHelper
      from_queue :testing_tasks

      def perform_work(msg, delivery_info, metadata)
        puts "I'm doing some task work in here yay!"
        return ResponseObject.success({test: true})
      end
    end
  end
end
```

The particular line this affects is: `@queue.channel.default_exchange.publish(response.to_reply, routing_key: metadata.reply_to, correlation_id: metadata.correlation_id)`

Which without this change can only access attr_reader :exchange which may be bind in such a way that this packet could not be delivered to the correct reply_to queue.

I hope this is a reasonable explanation!

By all means if there's a better way to do this, please let me know :+1: 

Thanks,
--Karl.


inventionlabsSydney:
@jondot Yo, can you look at why travis isn't allowing unparser?

````
Gem::InstallError: unparser requires Ruby version >= 2.1.
Installing nokogiri 1.6.7.2
An error occurred while installing unparser (0.2.5), and Bundler cannot
continue.
Make sure that `gem install unparser -v '0.2.5'` succeeds before bundling.
The command "eval bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}" failed. Retrying, 2 of 3.
```

Spaceghost:
Ruby 2.0.0 does not work with unparser. See my PR fix https://github.com/jondot/sneakers/pull/204/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R3.

gabrieljoelc:
I'm not sure exposing the channel is the right solution. Should we find a way to configure multiple publishers for a worker instead? Or can we define global publishers that we can configure in inilitializers?

Spaceghost:
@gabrieljoelc I am working on a bunny connection pool that publishers can check
connections in and out with.

If someone else has a better lead or just their own effort already, I'd love to
stand on your shoulders! I'm thinking about switching my custom pool out to use
the rather venerable [connection_pool][0]. I'm a huge fan of great interfaces
and find the simplicity comparable to a simpler [gen_server][1] for its domain.

[0]: https://github.com/mperham/connection_pool
[1]: http://erlang.org/doc/design_principles/gen_server_concepts.html

~Spaceghost

gabrieljoelc:
@Spaceghost the short answer is we use [`connection_pool`](https://github.com/mperham/connection_pool) for a global publisher (we call it the "bus"). We have rolled our own bus in our [la_gear](https://github.com/giftcardzen/la_gear) gem that uses a `Publisher` class that extends `Sneakers::Publisher`. Here's a [comment](https://github.com/FetLife/rollout/issues/49#issuecomment-150322755) that describes how we do it in detail (we really need to update our README to include something like that comment :grin:).

Spaceghost:
@gabrieljoelc I was thinking of supporting the pool directly within sneakers as
a configurable option. I could either pull request this to master or work on
getting some interfaces that would enable extension of sneakers.

What do you think?
~Spaceghost

lib/sneakers/version.rb
```diff
@@ -1,3 +1,3 @@
 module Sneakers
-  VERSION = "2.3.5"
+  VERSION = "2.3.6"
```
Spaceghost: Also, as a note, please don't ever do a version bump in a pull request. The
version designation is a responsibility of the release manager, whomever that
may be even if it's multiple persons or a robot.

Some projects include changelogs that include a ##next section for things on
develop that's not released that you would write the changelog notes into that
could help the release manager.

Hope this helps, cheers!

~Spaceghost

gabrieljoelc: :+1:

gabrieljoelc:
> work on getting some interfaces that would enable extension of sneakers.

@Spaceghost i'd love to see this over baking it in. I think it would be a more Sneaker-Way™ of doing things.

lib/sneakers/version.rb
inventionlabsSydney: :+1:  Understood! whoops!

gabrieljoelc:
Closing in favor of waiting for @Spaceghost to do [this](https://github.com/jondot/sneakers/pull/203#issuecomment-188899677). 😉

jacksonrayhamilton:
@gabrieljoelc / @Spaceghost, did [this](https://github.com/jondot/sneakers/pull/203#issuecomment-188899677) ever get done?  I'm hoping to implement an RPC with Sneakers like the one [here](https://www.rabbitmq.com/tutorials/tutorial-six-ruby.html).

michaelklishin:
@jacksonrayhamilton what specifically do you need to do? why do you need direct access to Sneakers' connection pool?

As of https://github.com/jondot/sneakers/commit/be5ee6bcfe9c4b9e0fc26fbc95853fbedd591b38, all  `Sneakers::Publisher` instances expose their channel. As of https://github.com/jondot/sneakers/commit/c38be02f95f93c18ce998340634edf59c8232ec7, the same is done for `Sneakers::Queue`. This makes it easy to publish using Bunny directly if you have to from both ends (publisher and consumer).

Sneakers is not necessarily designed for the request/response pattern but also doesn't prevent you from doing most of things with Bunny as long as you can get hold of the `Bunny::Channel` reference used internally. Publishing via Bunny in a way that makes sense for Sneakers workers is a matter of convention and understanding of [protocol basics](http://www.rabbitmq.com/tutorials/amqp-concepts.html) what Sneakers actually does with them.

jacksonrayhamilton:
@michaelklishin I'm needing to implement the functionality in the RPC call-receiving server from the tutorial, but wondering if I will be able to do that Sneakers, and perhaps if I can do it in a Sneakers worker (in our project, we have conventionally used Sneakers workers for receiving RMQ messages):

```rb
class FibonacciServer
  def start(queue_name)
    @q = @ch.queue(queue_name)
    @x = @ch.default_exchange

    @q.subscribe(:block => true) do |delivery_info, properties, payload|
      n = payload.to_i
      r = self.class.fib(n)

      @x.publish(r.to_s, :routing_key => properties.reply_to, :correlation_id => properties.correlation_id)
    end
  end
end
```

In my worker, I want to retrieve `properties.reply_to` and `properties.correlation_id`, and send a response back to the message originator.

One server needs another server to do some work.  The server that asked needs to wait until the working server is done.  Hence why I'm hoping to perform a [blocking] RPC.

michaelklishin:
@jacksonrayhamilton if a Sneakers worker responds to `:call_with_params`, it will be invoked instead of `:work`. It accepts the same parameters as `Bunny::Queue#subscribe` blocks but in a different order 😞: [payload, delivery_info, metadata](https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L58). From there what you can do what that data should be no different from Bunny ;)

jacksonrayhamilton:
Thanks @michaelklishin, this sounds like exactly what I need.

BTW, is the sneakers API documented anywhere?  For instance, is there a list of all the methods and how to use them?  I can only find some examples in the Wiki.  I feel like I might be more often able to self-answer these types of "how-tos" with a complete picture of the API.

michaelklishin:
@jacksonrayhamilton I don't think there's a reference. I [added it to rubydoc.info](http://www.rubydoc.info/github/jondot/sneakers/master) but documenting the API with YARD will take some time. Obviously we would consider contributions ;)

PR #204, opened 2016-02-17
Handle passing serverengine config in from runner

Spaceghost:
The `#run` method now accepts serverengine configuration options and
merges them against the server engine prior to the serverengine instance
having `#run` called on it.

Also support block acceptance on `#server_engine(&block)` which gives you
access to the runner that has `#server_engine` if you need to access that.
An internal, but an important one maybe worth sharing without too much
magic on the consumer side.

* Increased overall coverage & handled a couple TODOs which were solved
  by these changes.

```
  General coverage
     before: 91.78%
   new code: 91.34%
  new tests: 91.90%
      delta:  0.12%

  Coverage on lib/sneakers/runner.rb
     before: 73.81%
   new code: 72.00%
  new tests: 78.00%
      delta:  4.19%
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,5 @@
 require "bundler/gem_tasks"
-require 'metric_fu'
+require 'metric_fu' if RUBY_VERSION >= '2.1.0'
 require 'rake/testtask'
 
 Rake::TestTask.new do |t|
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,11 @@
 source 'https://rubygems.org'
 
+gem 'metric_fu' if RUBY_VERSION >= '2.1.0'
+gem 'simplecov'
+gem 'simplecov-rcov-text'
+gem 'guard'
+gem 'guard-minitest'
+gem 'nokogiri'
+gem 'ruby-prof'
+
 gemspec
--- a/lib/sneakers/runner.rb
+++ b/lib/sneakers/runner.rb
@@ -7,16 +7,31 @@ module Sneakers
       @runnerconfig = RunnerConfig.new(worker_classes, opts)
     end
 
-    def run
-      @se = ServerEngine.create(nil, WorkerGroup) { @runnerconfig.reload_config! }
-      @se.run
+    def run(config={}, &block)
+      self.server_engine = ServerEngine.create(nil, WorkerGroup) { @runnerconfig.reload_config!.merge!(config) }
+
+      server_engine(&block) if block_given?
+
+      server_engine.run
     end
 
     def stop
-      @se.stop
+      server_engine.stop
     end
-  end
 
+    def server_engine(&block)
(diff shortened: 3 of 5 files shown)


Rakefile
```diff
@@ -1,5 +1,5 @@
 require "bundler/gem_tasks"
-require 'metric_fu'
+require 'metric_fu' if RUBY_VERSION >= '2.1.0'
```
gabrieljoelc: I see you have the version condition in the `Gemfile`, but should this be more of a `defined?` condition here?

gabrieljoelc:
I think this is an excellent and gives more control over the concurrency. Is this the direction you want to go or do you want some of this to be handled by [`sneakers.conf.rb`](https://github.com/jondot/sneakers/wiki/Configuration#auto-scaling-configuration)?

Rakefile
Spaceghost: What would you pass to defined in this case?

```ruby
defined?(RUBY_VERSION) # => "constant"
```

gabrieljoelc: I think I'm stupid. Nothing is loaded to pass to define? yet. ☺️

Spaceghost:
@gabrieljoelc I just wanted to ping you to let you know I'm braining on your
[last question][0]. I think that it merits a thought out response, since it
seems to cross public/internal/private API boundaries in ways that we might
need to really own later and not drop if we expose them publicly.

Sorry for the delay, and thank you for the review! If any part of this might be
easier to initially discuss in some other way (hangouts, screenhero, IRC,
    whatever) and then recapture here later, I'd love to chat or pair.

I have a few more bunny-related pull requests incoming for some fixes, features
(TLS mutual auth and friends support? You betcha.), and then some
architectural stuff that might even be more worthy of discourse. I'll get back
with my digested answer to your question.

[0]: https://github.com/jondot/sneakers/pull/204#issuecomment-188633123

~Spaceghost

Rakefile
Spaceghost: No worries, I was more excited to learn some black magic ruby spell that I
don't actually know yet!

I'm not a huge fan of actually using `require 'bundler/setup'` and all of that,
especially inside gems, I prefer to require inside library code not indirectly
through the Gemfile via bundler. I hope that makes sense.

~Spaceghost

Spaceghost:
@arfrank I think the most correct thing to add after squashing the last two would be `~> 2.2` unless there's some features that are depended on in versions greater than 2.2.

That would be applicable for everyone, and if your temporary version doesn't match into the `2.x.y.alpha` pattern then you'd maybe want to use your own version of them gem installed whichever way works best into you workflow.

michaelklishin:
@Spaceghost can you please rebase this against master?

Spaceghost:
@michaelklishin I'll have to close this one and open one from my own github account. I'll update this with the new pull request.

PR #205, opened 2016-02-18
Add another maxretry handler which does not create additional exchanges

nevesenin:


Spaceghost:
I bet to get this merged, it'll be helpful to convert from rspec back into minitest like. If you'd like some help with this, I could either pair with you on it or I might be able to take a look and try to convert it myself. If you're not merged in by then, I'll pull request to your branch.

nevesenin:
Hi, I don't have much experience with minitest. To be honest, I don't know what you mean.

lib/sneakers/handlers/routing_maxretry.rb
```diff
@@ -0,0 +1,207 @@
...
+
+      def reject_to_retry(delivery_info, message_properties, num_attempts)
+        Sneakers.logger.info do
+          "#{log_prefix} msg=retrying, count=#{num_attempts}, headers=#{message_properties.headers}"
+        end
+
+        channel.reject(delivery_info.delivery_tag)
+      end
+
+      def error_payload(delivery_info, message_properties, payload, reason, num_attempts)
```
adisos: Could you explain why we need to change original payload?

nevesenin: To add debug information in case of an error.

JanStevens:
Cool, how does this differ in terms of the existing max retry exchange?
It does not create extra exchanges but does this have a performance penalty? Any change you can draw some topology out?

thanks looks interesting!

JanStevens:
After some experimenting a couple of pointers for other people that want to try:

when making a worker you must add the following arguments:
```ruby
class Worker
  from_queue :foo, handler: Sneakers::Handlers::RoutingMaxretry,
                              arguments: { 'x-dead-letter-exchange' => "exchange",
                                                   'x-dead-letter-routing-key' => "queue.foo.retry" }

  # ....
end
```
Notice that its `.retry` and not `-retry` as with the current maxretry handler

Next, wildcard route matching is quite dangerous. I had a general routing key from my `exchange` to `foo` using `#` but this creates an endless feedback loop. If any of your defined routing_keys also match the following then you might have problems (I'm using example queue name `foo`):

- `queue.foo.error`
- `queue.foo.requeue`
- `queue.foo.retry`

TheBerg:
I am working on using this, but when an error goes to the error queue bunny errors and gives me a `IOError: not opened for reading` error. I am investigating but just thought I would post here if anyone has got an idea. :)

PR #210, opened 2016-03-22
Allow to pass Proc as worker classes

sharshenov:
This small PR gives ability to configure worker list in more flexible way.
Here is a pseudo-code usage example:

I. Put workers list in `workers.yaml`
```yaml
- FooWorker
```

II. [Run Sneakers programmatically](https://github.com/jondot/sneakers/wiki/How-to:-running-a-stand-alone-worker#running-sneakers-programmatically)
```
r = Sneakers::Runner.new -> { YAML.load_file('workers.yml').map(&:constantize) }
r.run
```

III. Edit `workers.yaml`
```yaml
- FooWorker
- BarWorker
```

IV. Run ```$ kill -SIGUSR1 `cat sneakers.pid` ``` to [gracefully restart sneakers](https://github.com/fluent/serverengine#signals).

### DIFF
--- a/lib/sneakers/workergroup.rb
+++ b/lib/sneakers/workergroup.rb
@@ -23,7 +23,13 @@ module Sneakers
       # when used with many workers.
       pool = config[:share_threads] ? Thread.pool(config[:threads]) : nil
 
-      @workers = config[:worker_classes].map{|w| w.new(nil, pool) }
+      worker_classes = config[:worker_classes]
+
+      if worker_classes.respond_to? :call
+        worker_classes = worker_classes.call
+      end
+
+      @workers = worker_classes.map{|w| w.new(nil, pool) }
       # if more than one worker this should be per worker
       # accumulate clients and consumers as well
       @workers.each do |worker|


gabrieljoelc:
So do you have some workers in your codebase you don't want to booted as subscribers and so you want to be able to whitelist the workers you want running?

sharshenov:
No. I want to have ability to turn on/off workers on-the-fly.

gabrieljoelc:
This will require redeployment to toggle them whereas ENV change can just be a restart. Does that matter?

sharshenov:
No redeployment is necessary. Only USR1 signal to supervisor process. After that supervisor process [will not be restarted but will restart child processes](https://github.com/fluent/serverengine#signals). 
Getting worker list from yaml file is just example. You can get it from whatever you want(worker that reconfigures sneakers for example).

gabrieljoelc:
oh yea, cool!

michaelklishin:
Not sure how many would use this but it looks reasonable, thank you!

sharshenov:
After 2 years since I suggested this feature I've finally started using it 🐌
[Laces for sneakers](https://github.com/sharshenov/sneakers-laces) allows to declare queues on producer side and gracefully restart consumer to subscribe to them without redeployment.

PR #217, opened 2016-05-12
include headers in Maxretry Handler error messages, plus a rake task to retry failed messages

graimon:
First of all, thanks for your work.

While I was creating a task to requeue failed tasks, I realized there was no information about the headers in the error queue message. It's been easy to add it :)

Also, I've thought it may be useful to share the task with everyone.

### DIFF
--- a/lib/sneakers/tasks.rb
+++ b/lib/sneakers/tasks.rb
@@ -62,15 +62,22 @@ EOF
     end
 
     worker_class = ENV['WORKER'].constantize
-    config = Sneakers::CONFIG.merge worker_class.queue_opts
+    
+    config = Sneakers::CONFIG.merge(worker_class.queue_opts)
+
+    durable = config.fetch(:queue_options, {}).fetch(:durable, false)
+    requeue_exchange = config[:retry_requeue_exchange] || worker_class.queue_name + "-retry-requeue"
+    error_queue = config[:retry_error_exchange] || worker_class.queue_name + "-error"
+
     bunny = Bunny.new(config[:amqp], :vhost => config[:vhost], :heartbeat => config[:heartbeat], :logger => Sneakers::logger)
     bunny.start
     channel = bunny.create_channel
     channel.prefetch config[:prefetch]
-    exchange = channel.exchange(config[:exchange], config[:exchange_options])
-    queue = channel.queue(worker_class.queue_name + "-error", :durable => true)
 
-    delivery_info, properties, payload = queue.pop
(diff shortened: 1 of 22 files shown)


lib/sneakers/tasks.rb
```diff
@@ -41,4 +41,47 @@
...
+    worker_class = ENV['WORKER'].constantize
+    config = Sneakers::CONFIG.merge worker_class.queue_opts
+    bunny = Bunny.new(config[:amqp], :vhost => config[:vhost], :heartbeat => config[:heartbeat], :logger => Sneakers::logger)
+    bunny.start
+    channel = bunny.create_channel
+    channel.prefetch config[:prefetch]
+    exchange = channel.exchange(config[:exchange], config[:exchange_options])
+    queue = channel.queue(worker_class.queue_name + "-error", :durable => true)
+
+    delivery_info, properties, payload = queue.pop
```
justin-yesware: Popping the message off the queue also acks it. So if this rake task fails between this line and when it's published below, this message is lost.

lib/sneakers/tasks.rb
```diff
@@ -41,4 +41,47 @@
...
+
+    while payload
+      error_data = JSON.parse(payload)
+      msg = Base64.decode64(error_data["payload"])
+      
+      exchange.publish msg, :routing_key => delivery_info[:routing_key], :headers => error_data["headers"]
```
justin-yesware: I believe this will publish it back to the original exchange, which may have multiple bindings. This has the effect of re-broadcasting this message to all consumers if only a single consumer fails to process it. You should publish messages to the `worker_class.queue_name + "-retry-requeue"` exchange which is setup to have a single binding, the worker's queue for just this purpose.

lib/sneakers/tasks.rb
graimon: Thanks for the tip! I didn't have that in mind. I'll change it and push again.

lib/sneakers/tasks.rb
graimon: And again, thank you. Publishing to the original exchange was enough for our needs, but I see the implications. I will also change it.

lib/sneakers/tasks.rb
michaelklishin: You can use manual acknowledgements with `basic.get` (`Bunny::Queue#pop`).

michaelklishin:
We now include all properties. I believe this no longer applies. Let me know if I'm wrong.

Issue #171, opened 2015-10-06
Sneakers should warn if it was run without configuration first


jondot:
Related to #170, either configure with defaults always, or warn users and suggest initializer or a snippet of configuration code.

Issue #170, opened 2015-10-05
rake sneakers:run fails with NoMethodError: private method `warn' called for nil:NilClass


selasiehanson:
Hello, 
I'm not sure if I am doing something wrong but in my rails app, anytime I run `WORKERS=ImageConverter rake sneakers:run`
I come up with the following error. I had the same issue in 2.2.1. I upgraded the gem to 2.3.0 and still the error persists. Any help will be appreciated.

 :/web_app# WORKERS=ImageConverter rake sneakers:run
rake aborted!
NoMethodError: private method `warn' called for nil:NilClass
/usr/local/bundle/gems/sneakers-2.3.0/lib/sneakers/runner.rb:43:in `reload_config!'
/usr/local/bundle/gems/sneakers-2.3.0/lib/sneakers/runner.rb:11:in `block in run'
/usr/local/bundle/gems/serverengine-1.5.11/lib/serverengine/config_loader.rb:39:in `call'
/usr/local/bundle/gems/serverengine-1.5.11/lib/serverengine/config_loader.rb:39:in `reload_config'
/usr/local/bundle/gems/serverengine-1.5.11/lib/serverengine/config_loader.rb:32:in `initialize'
/usr/local/bundle/gems/serverengine-1.5.11/lib/serverengine/daemon.rb:29:in `initialize'
/usr/local/bundle/gems/serverengine-1.5.11/lib/serverengine.rb:51:in `new'
/usr/local/bundle/gems/serverengine-1.5.11/lib/serverengine.rb:51:in `create'
/usr/local/bundle/gems/sneakers-2.3.0/lib/sneakers/runner.rb:11:in `run'
/usr/local/bundle/gems/sneakers-2.3.0/lib/sneakers/tasks.rb:38:in `block (2 levels) in <top (required)>'
Tasks: TOP => sneakers:run
(See full trace by running task with --trace)

jondot:
I think this is due to not running Sneakers.configure anywhere. Create an initializer script in your Rails app and in it configure Sneakers like mentioned below:
https://github.com/jondot/sneakers/wiki/How-To:-Rails-Background-Jobs#global-configuration

Please let me know if that helps

selasiehanson:
@jondot That works. Thanks a lot. I put the sneakers.rb in the config folder instead of config/initializers

PR #220, opened 2016-06-06
Log to stdout by default

goromlagche:
This is in response to https://github.com/jondot/sneakers/issues/171

> Related to #170, either configure with defaults always, or 
> warn users and suggest initializer or a snippet of configuration code.

What I have found is, `sneakers` already has default configuration, its just the `logger` does not gets initiated, so I think preferable thing would be to `configure with defaults always` rather than `warn users and suggest initializer or a snippet of configuration code.`

Another usecase I have in favour of adding to default configuration is this - 
I configure sneakers with an initializer except for the test env

```
Sneakers.configure(Rails.configuration.x.sneakers) unless Rails.env.test?
```

Now I use `Sneakers::Publisher` to publish to queue. And to write specs, I use `bunny-mock` to create a mock exchange and stub it. 

```
publisher = Sneakers::Publisher.new(exchange: 'athena_portal',
                                        exchange_options: { durable: true })
publisher.instance_variable_set(:@exchange, mock_exchange)
```

And then publish to mocked queue and check the content of queue to see if `Sneakers::Publisher` publishes properly or not. Now `publisher` has this code https://github.com/jondot/sneakers/blob/master/lib/sneakers/publisher.rb#L14

```
Sneakers.logger.info {"publishing <#{msg}> to [#{options[:routing_key]}]"}
```

Now just for this line I have to write 

```
logger = Logger.new('log/sneakers.log')
Sneakers.configure(log: logger)
```

if Sneakers had a default `logger` I wouldn't have to do this.

p.s. I dont have much knowledge on `sneakers` or `rabbitmq` I just been using these for last couple of weeks. So to add a default logger I just debugged for a while and added some code. Please let me know if we can do these following changes in a better way, or should we do this at all.

### DIFF
--- a/lib/sneakers/concerns/logging.rb
+++ b/lib/sneakers/concerns/logging.rb
@@ -10,7 +10,7 @@ module Sneakers
 
       module ClassMethods
         def logger
-          @logger ||= Logger.new(STDOUT)
+          @logger ||= configure_logger
         end
 
         def logger=(logger)
@@ -25,6 +25,7 @@ module Sneakers
             @logger.level = Logger::INFO
             @logger.formatter = Sneakers::Support::ProductionFormatter
           end
+          @logger
         end
       end
     end
--- a/spec/sneakers/concerns/logging_spec.rb
+++ b/spec/sneakers/concerns/logging_spec.rb
@@ -9,19 +9,19 @@ end
 
 describe Sneakers::Concerns::Logging do
   describe ".configure" do
-    before do
-      Foobar.logger = nil
+
+    it "should configure a default logger" do
+      Foobar.logger.wont_be_nil
+      Foobar.logger.formatter.must_equal Sneakers::Support::ProductionFormatter
     end
 
     it "should configure a default logger when included" do
-      Foobar.logger.must_be_nil
       Foobar.configure_logger
       Foobar.logger.wont_be_nil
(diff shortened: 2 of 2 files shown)


michaelklishin:
@goromlagche can you please see if this is still relevant and rebase against master?

michaelklishin:
@gabrieljoelc WDYT about this idea in general? I think the "log everything to standard streams" camp is seriously overstating the number of environments that do that and it's quite easy to configure the logger the way they need anyway.

On the other hand, there is no good default for file-based logging even if we ignore Windows for a moment.

gabrieljoelc:
@michaelklishin Not everyone uses sneakers with Rails, but it seems reasonable to go with the convention of defaulting to file logging because Rails does.

michaelklishin:
I think we will keep the default we have for now. Plenty (I'd argue the majority) of environments do not log to standard streams and if you have to, it's easy to do that.

alilland:
i landed on this issue because I have no idea how to get logging to STDOUT in development, so +1 for OP

michaelklishin:
@alilland you should be able to set the logger [using `Sneakers.logger=`](https://github.com/jondot/sneakers/blob/31d0cb25dc5bbcfb0749567e9e0f80e6353fb66b/lib/sneakers/concerns/logging.rb) to anything you need.

PR #223, opened 2016-06-20
Ensure queue exists before publishing to avoid lost messages

gideonaa:
Publishing to a queue that doesn't exist (or you could say an exchange that has no queues bound to it) will silently push the message into the cyber abyss. This can lead to an ugly data loss.

I think the publisher class should make sure and declare the queue to which it's publishing.  This way if the queues on the broker ever disappear (say in the event of a server crash/restart), our publisher will not naively publish a message into /dev/null.

### DIFF
--- a/lib/sneakers/configuration.rb
+++ b/lib/sneakers/configuration.rb
@@ -21,6 +21,10 @@ module Sneakers
       :arguments => {}
     }.freeze
 
+    PUBLISHER_OPTION_DEFAULTS = {
+      :auto_declare_queue => true
+    }
+
     DEFAULTS = {
       # Set up default handler which just logs the error.
       # Remove this in production if you don't want sensitive data logged.
@@ -46,7 +50,8 @@ module Sneakers
       :hooks              => {},
       :exchange           => 'sneakers',
       :exchange_options   => EXCHANGE_OPTION_DEFAULTS,
-      :queue_options      => QUEUE_OPTION_DEFAULTS
+      :queue_options      => QUEUE_OPTION_DEFAULTS,
+      :publisher_options  => PUBLISHER_OPTION_DEFAULTS
     }.freeze
 
 
--- a/lib/sneakers/publisher.rb
+++ b/lib/sneakers/publisher.rb
@@ -8,7 +8,7 @@ module Sneakers
     def publish(msg, options = {})
       @mutex.synchronize do
         ensure_connection! unless connected?
-        ensure_queue!(options[:to_queue])
+        auto_declare_queue(options[:to_queue]) unless !@opts[:publisher_options][:auto_declare_queue]
       end
       to_queue = options.delete(:to_queue)
       options[:routing_key] ||= to_queue
@@ -38,12 +38,12 @@ module Sneakers
       Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
     end
 
-    def ensure_queue!(name)
-      # If the @queue attribute is already set, simply return
-      # otherwise declare+bind queue to ensure the msg is sent somewhere
+    def auto_declare_queue(queue_name)
+      # Return if @queue attribute already
+      # otherwise declare & bind queue to ensure the msg is sent somewhere
       return if @queue
-      @queue = @channel.queue(name, @opts[:queue_options])
-      @queue.bind(@exchange, routing_key: name)
+      @queue = @channel.queue(queue_name, @opts[:queue_options])
+      @queue.bind(@exchange, routing_key: queue_name)
     end
   end
 end
--- a/lib/sneakers/publisher.rb
+++ b/lib/sneakers/publisher.rb
@@ -8,6 +8,7 @@ module Sneakers
     def publish(msg, options = {})
       @mutex.synchronize do
         ensure_connection! unless connected?
+        ensure_queue!(options[:to_queue])
       end
       to_queue = options.delete(:to_queue)
       options[:routing_key] ||= to_queue
@@ -16,7 +17,7 @@ module Sneakers
     end
 
 
-    attr_reader :exchange
+    attr_reader :exchange, :queue
 
   private
     def ensure_connection!
@@ -35,6 +36,14 @@ module Sneakers
 
     def create_bunny_connection
       Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
+    end
+
+    def ensure_queue!(name)
+      # If the @queue attribute is already set, simply return
+      # otherwise declare+bind queue to ensure the msg is sent somewhere
+      return if @queue
+      @queue = @channel.queue(name, @opts[:queue_options])
+      @queue.bind(@exchange, routing_key: name)
     end
   end
 end
--- a/spec/sneakers/publisher_spec.rb
+++ b/spec/sneakers/publisher_spec.rb
@@ -52,7 +52,7 @@ describe Sneakers::Publisher do
       p = Sneakers::Publisher.new
       p.instance_variable_set(:@exchange, xchg)
       p.instance_variable_set(:@queue, xchg)
-
+ 
       mock(p).ensure_connection! {}
       p.publish('test msg', to_queue: 'downloads', expiration: 1, headers: {foo: 'bar'})
     end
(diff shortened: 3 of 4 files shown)


sharshenov:
As I know, this is an expected behaviour. RMQ router is responsible for making decisions about queues & message traffic. So publisher should do the only thing - publish and nothing more.

gideonaa:
@sharshenov I had the same concern at first but since the publisher goes as far as to declare the exchange I didn't think this would be a big scope change. 

We could write a setup task to create the queues, exchanges and bindings (like I have on my project) but it's tricky to make sure this definitely runs in the event that the broker dies and the queues disappear. Just seems like a few lines that could be a life-saver if you're unlucky enough to have lost your queues without knowing it.

gideonaa:
Any final thoughts on this? @jondot ? Been using this branch in my project and it's a nice broker-restart-proof publisher.

If we have consensus I can start working through the spec failures.  Thanks!

jondot:
Hey @gideonaa, so I hope to give you some context, here goes.

Publisher was always a second-class thing in sneakers, as opposed to a worker which is first-class. This is because publishing involves much less 'effort' than consuming and actually doing work. For this reason, publishers a lot of times tend to be not Ruby at all, and sometimes it even makes more sense to have a publisher in Go, Node, or any system that can handle performance in a better way than Ruby.

This is why, originally, you don't see these things such as the one you just fixed addressed in the first place, although critical things like concurrency were addressed in Publisher from the get-go in order to not have people take risks.

Now why wasn't any queue creation done in the publisher if one didn't exist? to answer this question we need to first answer - "who owns a queue"?

You have 3 options:

1. The publisher
2. The consumer (worker)
3. A 3rd party (not the publisher nor the consumer)

The short answer is that since Sneakers is a background worker library, is owner is the worker (1). This means the worker goes up first, creates the queue and setup the exchange. If there were never any workers for a certain setup, then this setup would not exist, then there is no point in pushing messages to that set up.

Where this breaks, is in pub-sub topologies, where there can be several types of publishers, several types of consumers, all looking at the same queue. In that scenario - who "owns" the queue? the answer is - no one. You will need to have some set up code to prepare the queues for both publisher and consumer. This scenario can be seen in a more harsh ways on Kafka setups.

When does a publisher "owns" a queue? I have yet to come at scenario that can frame only a publisher to own a queue.

So, this sums up the context for the publisher story. I hope it makes sense in the right places.

Going back to your idea - as I finished working on the publisher, I had that fear in my mind ("what if someone will make a publisher without a consumer? will those be lost? will he know?") and I decided, since I saw no real use cases, that I'll have to meet that someone first.
So - hello! :)
In that sense, I welcome this addition, since clearly you have a use case that moved you to solve this. I'd like to hear more about it for academic reasons. 

I think what would be perfect is to encourage to have this functionality opt-in (such that a user can specify if he wants a publisher to define the queue/exchange via options). If you can see this feature going that way then I would be excited to have this merged in.

gideonaa:
Thanks a lot for the response @jondot, I totally get it.  The worker framework is definitely what brought me to sneakers but the publisher class provided is neat enough to use on the pub side as well.

As far as my application goes it's closest to the pub-sub topology where there can be multiple publishers pushing to the same queue.  The best way to describe our use is that we have an ETL project where the extract/transform part happens in multiple different projects which end up pushing their results to the RMQ broker and the load part happens on the workers side.  So the publishers are separate projects that find data (either through partner APIs, web crawling, etc) and then push them to the queue in a fire-and-forget manner. The workers are a separately managed project with the sole focus of grabbing those messages and efficiently committing them to storage (along with a bit of pre-processing).  So while our workers create queues before they start consuming, the publishers run on their own separate schedule and can't assume that a queue exists. So basically, our publishers are mostly the owners but our workers are critically important.

ANYWAY, I like the idea of making it optional! Maybe we can have a `publisher_options` section in the `Sneakers.config` with an option like `ensure_queues`?

jondot:
@gideonaa thanks, sorry for the late response, I've been flying around last week.
I secretly guessed it was a many publishers to many consumers topology, happy to see it validated :)
So, definitely, I think it is a good step to recognize an owner - which you have (publishers). I like the idea of `publisher_options` as well, and regarding naming - I'd say `ensure` doesn't tell me if it will automatically create: will it fail after not being able to ensure? will it create and continue? so in that sense I guess minimizing the element of surprise would be something that would hint this creates queues if missing - makes sense?

gideonaa:
@jondot no worries, I appreciate the level of activity in this project, it's great. 

Very good point about the naming.  So I'm thinking we name the option `auto_declare_queue` which will reside in the `publisher_options` config.  What do you think?  I'm glad there's someone out there who's as serious about naming 👍

jondot:
Sounds good! `auto_declare_queue` is clear to me - there's no surprise about what it does :)
👍 👍

michaelklishin:
This has the same downside as #193. Unless there is a way to pre-declare queues on start, I strongly suggest not accepting this PR.

gideonaa:
@michaelklishin my understanding of the downside in #193 is that the `with_channel` creates a temporary channel per message whereas I'm trying to declare the queue/bindings only if it hasn't already been declared (and if the `auto_declare_queue` config is set). 

As far as I can see the only extra action per message will be evaluating `if @queue` which shouldn't slow down messaging substantially. Maybe I'm missing something though 😕 .

Anyways, I'll hold off on writing tests and making the proposed `auto_declare_queue` setting until I hear back.

michaelklishin:
@gideonaa ok, I can see an early termination condition now. This looks better.

gideonaa:
@jondot alrighty, I think that covers everything discussed so far.  Let me know if there's anything else you'd like to see before it's merge-able.

gideonaa:
@jondot just checking back on this.  Would love to merge so I can start using the master branch in my project!  Thanks in advance.

gabrieljoelc:
Sorry I'm late to this discussion, but I do have one thought related to our use case. Our publishing and subscribing code will often times be in two different applications/repos, respectively. For deployment, we end up labeling the PR in the publishing repo as "blocked" until the PR in the subscribing repo is merged and deployed. The queue in the subscribing application is namespaced to that application and bound to the direct exchange with a routing key that is _not_ namespaced.

`my_subscribing_app.my_queue_name` is bound to the `my_direct_exchange` with `my_routing_key`.

With this use case, the publishing app would have to know about the subscribing app's convention to create it ahead of time, defeating the purpose of de-coupled pub-sub applications.

michaelklishin:
@gabrieljoelc good point!

gideonaa:
@gabrieljoelc I totally see your point and it's not far from what @jondot was talking about in his comment regarding 'who owns the queue'.  That's why we decided to make this configuration optional: https://github.com/jondot/sneakers/pull/223/files#diff-ee4ef6a4f02dcd635c5738304fa04d0fR25

But maybe I ought to default that config to false?

As a side note, I'd say it doesn't remove the benefits of a decoupled app it just guarantees that the messages will be sent somewhere in the event that whatever app builds 'messaging fabric' fails to declare/bind a queue to the exchange.  So I see it more as a failsafe. But you're right that it's mostly useful when your subscribing applications can simply use the default routing key (i.e; the queue name), as is the case with me.

gabrieljoelc:
The general ambiguity of the direction of sneakers could play into a decision on a feature like this. I would like to see sneakers core stay more general, and opinionated or use-case-specific features like this be sneakers plugins in separate repos that could be registered as "middleware" or something. I would rather see work done on making sneakers more extendable than adding more configuration options added. 😬 

> But maybe I ought to default that config to false?

If enough people want this in the meantime, I am willing to compromise with this default.

sharshenov:
+1 to keep sneakers simple.
I also vote for extracting publisher to a separate gem. That will make sneakers to work with consuming only and to stick to RabbitMQ-way.

michaelklishin:
I completely agree with @gabrieljoelc. Note that the definition of "simplicity" tends to vary from person to person but I suspect that most developers would agree that if we can keep additional non-essential (according to the core team) features in plugins or side projects to Sneakers, it would likely keep Sneakers itself smaller and thus simpler.

gabrieljoelc:
@gideonaa thanks for your interest in sneakers and for contributing. Also, thank you for triggering us to firm our roadmap for sneakers core.

If you interesting in helping make sneakers more extendible and creating extensions for it, please contact one of us by email to continue the discussion.

Closing for now.

gideonaa:
@gabrieljoelc sure thing, I wouldn't want to take the project in a direction the group doesn't want or overcomplicate things.  Since the Publisher was already declaring the exchanges, I didn't think it was a big change of scope.

Separating the Publisher to a different gem like @sharshenov could help keep things simple or some plugin/middleware could be started for spinning up the messaging fabric (i.e; bootstrapping tool for declaring queues, exchanges, bindings, clusters, etc).  Don't have the time for that kind of refactor now but maybe I'll do it in the future 🎊

PR #233, opened 2016-07-21
Fix broken specs

pbollenbeck:
Specs have been failing for a while. These changes fix the broken specs and introduce a travis build status badge in the readme.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ language: ruby
 rvm:
 - ruby-head
 - 2.2
-- 2.1
-- 2.0.0
 matrix:
   include:
     - rvm: 2.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,10 @@ cache: bundler
 language: ruby
 rvm:
 - ruby-head
-- 2.2
+- 2.2.3
 matrix:
   include:
-    - rvm: 2.2
+    - rvm: 2.2.3
       env: INTEGRATION_LOG=1 INTEGRATION=1
 
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Sneakers
 
+[![Build Status](https://travis-ci.org/jondot/sneakers.svg?branch=master)](https://travis-ci.org/jondot/sneakers)
 
 ```
       __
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -6,12 +6,10 @@ cache: bundler
 language: ruby
 rvm:
 - ruby-head
-- 2.2
-- 2.1
-- 2.0.0
```
Spaceghost: Tests might pass, but is dropping these rubies the right thing? I'm fine with either way, but sneakers has less viability when it only cares to support the latest stable.

pbollenbeck: The problem lies in the development dependencies. We could remove at least `metric_fu` which depends on `activesupport` and `rails_best_practices` which have pretty strict ruby requirements themselves. `metric_fu` doesn't seem to be used in the CI as far as I have gone through the code (but I'm not too familiar with it yet).
I'd in general favor a lighter list of dev dependencies anyway, IMO also `simplecov`, `simplecov-rcov-text`, `guard`, `guard-minitest`, `ruby-prof` and `metric_fu` could be removed without harm.

Spaceghost: See #204, where I fixed that.

pbollenbeck: @jondot any preference?

michaelklishin:
👍

michaelklishin:
I'll stick my neck out and merge this because having functional CI is very important. @jondot I hope requiring Ruby 2.1+ is reasonable. Some of the changes can be undone if you disagree.

gabrieljoelc:
from https://github.com/jondot/sneakers#compatibility:
> * Sneakers 1.1.x and up (using the new generation Bunny 2.x) - Ruby 2.x.x
> * Sneakers 1.x.x and down - Ruby 1.9.x, 2.x.x

Do we even know what was specific to Ruby `2.0.0`?

jondot:
There were some specific requirement for Ruby 1.9 but I believe there's no need to support these anymore. 2.x and up should be fine :thumbsup:

jondot:
Also - this means 2.1.x and up is good to go!

README.md
```diff
@@ -1,5 +1,6 @@
 # Sneakers
 
+[![Build Status](https://travis-ci.org/jondot/sneakers.svg?branch=master)](https://travis-ci.org/jondot/sneakers)
```
kruppel: @pbollenbeck Just in case is was unintentional, this is the second badge added to the README (the first is on line 23) 😄.

pbollenbeck: Oh, you're totally right. My apologies, I missed this. As the PR is already closed, how should we proceed? New PR?

PR #234, opened 2016-07-21
Disable Rake::TestTask warnings

pbollenbeck:
This will avoid noisy tests with lots of warnings introduced with
ruby/rake@8078bb6

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,6 +5,7 @@ require 'rake/testtask'
 Rake::TestTask.new do |t|
   t.libs << "spec"
   t.test_files = FileList['spec/**/*_spec.rb']
+  t.warning = false
 end
 
 task default: :test


pbollenbeck:
I'd suggest to merge #233 first to make this pass travis, but I don't know what the PR/merge habits/requirements/policies in this repos are right now.

Spaceghost:
I don't support libraries suppressing warnings. I'd rather have noisy tests myself. If you need to quiet the log output for your mind, write a logger or change the file for yourself, I don't believe a library benefits from hiding warnings.

pbollenbeck:
This commit just reinstates the behavior of rake before ruby/rake@8078bb6. The library itself doesn't suppress warnings, it's just the warnings thrown by code of (older) gems sneakers is depending on. Warnings aren't disabled when sneakers is used in production code.

pbollenbeck:
BTW, the warnings are mainly coming from: `amq-protocol` (2.0.1), `bunny` (2.5.0), `minitest` (5.9.0), `guard` (2.14.0). In addition, there's a bunch of warnings coming from spec/sneakers/worker_handlers_spec.rb:49 (`instance variable @queue not initialized`)

PR #235, opened 2016-07-21
Relax bunny version requirement

pbollenbeck:
Current version of bunny (2.5.0) includes a lot of bug fixes relevant for
most production environments. This commit makes sneakers depending on
bunny '~> 2.2' and therefore allow all 2.x versions.

PR #236, opened 2016-07-22
Bump Bunny to 2.5.0

michaelklishin:


pbollenbeck:
IMO it's better to just relax the version requirement to 2.x based on semantic versioning (see #235). There is no need to update the version requirement with every new minor bunny version.

PR #257, opened 2016-11-30
Pass `routing_key` to `Sneakers::Worker.publish` if available

kruppel:
Routing key can be defined via `.from_queue` and should be respected when publishing via the class method.

### DIFF
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -131,8 +131,11 @@ module Sneakers
         @queue_opts = opts
       end
 
-      def enqueue(msg)
-        publisher.publish(msg, :to_queue => @queue_name)
+      def enqueue(msg, opts={})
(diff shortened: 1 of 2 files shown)


gabrieljoelc:
Thanks for contributing!

Looks interesting. How are you using this right now?

Can you add some tests please?

kruppel:
> Looks interesting. How are you using this right now?

We're actually not using Sneakers to publish messages. We have a homegrown solution for doing this sort of thing. However, it's generic enough where I'd like to move to using Sneakers for both subscription and publishing.

> Can you add some tests please?

Done! I reduced the existing tests for `.enqueue` to a single test because they were all doing basically the same thing. Let me know what you think.

Thanks for the review!

michaelklishin:
Thank you!

PR #260, opened 2016-12-21
Add sneakers-specific timeout error

### LINKED ISSUE #259: Introduce a specific error to replace Timeout::Error
While `Timeout::Error` speaks for itself, it can be ambiguous as well. Being part of the standard Ruby library it is available for various gems and for the Ruby library itself. That and the fact sneakers jobs are authored with long-running gem/stdlib/ruby code to do some work it can easily become ambiguous in figuring out which part of the code raised the error - the sneakers gem or the long-running code inside the job which in my case also has a `Timeout.timeout` piece of code. 

Additional benefit of this change is that it will improve error reporting/searching based on class name, seeing 100 occurrences of `Timeout::Error` is less valuable than 50 Sneakers-specific error and 50 `Timeout::Error`.

There is an expectation, which may not be shared by everyone, that when a library would like to raise an error it should be prefixed with library's namespace to be explicit in its meaning, purpose. Looking it from the opposite angle, seeing a library raising `StandardError`, `RuntimeError` (`Timeout::Error`'s parent) doesn't tell me much which means taking a closer look at error messages, stacktraces and such. I have took a look at most popular gems (bundler, rake, rack) and they all define some prefixed errors. This article has some good points on the topic http://weblog.jamisbuck.org/2007/3/7/raising-the-right-exception.html

To summarize, seeing a `Sneakers::TimeoutError/Sneakers::Worker::TimeoutError` will be of much more help than the common `Timeout::Error`.
Thank you for this gem!

dsbonev:
This distinguishes sneakers job timeouts from timeouts raised by
job code (e.g. an http lib with connection timeout).

More details are available in the related issue.
Fixes #259

### DIFF
--- a/lib/sneakers/errors.rb
+++ b/lib/sneakers/errors.rb
@@ -1,0 +1,5 @@
+require 'timeout'
+
+module Sneakers
+  class WorkerTimeout < Timeout::Error; end
+end
--- a/lib/sneakers.rb
+++ b/lib/sneakers.rb
@@ -12,6 +12,7 @@ module Sneakers
 end
 
 require 'sneakers/configuration'
+require 'sneakers/errors'
 require 'sneakers/support/production_formatter'
 require 'sneakers/concerns/logging'
 require 'sneakers/concerns/metrics'
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -51,7 +51,7 @@ module Sneakers
 
         begin
           metrics.increment("work.#{self.class.name}.started")
-          Timeout.timeout(@timeout_after, Timeout::Error) do
+          Timeout.timeout(@timeout_after, WorkerTimeout) do
             metrics.timing("work.#{self.class.name}.time") do
               if @call_with_params
                 res = work_with_params(msg, delivery_info, metadata)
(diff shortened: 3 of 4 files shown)


lib/sneakers/errors.rb
```diff
@@ -0,0 +1,5 @@
+require 'timeout'
+
+module Sneakers
+  class JobTimeout < Timeout::Error; end
```
gabrieljoelc: Would it more aligned if this was `WorkerTimeout`?

michaelklishin: Or even `Sneakers::Timeout`.

dsbonev: > Would it more aligned if this was WorkerTimeout?

Done

dsbonev: > Or even Sneakers::Timeout.

I did try that as well, but 29 specs failed. I see the last one is about
> +"private method `timeout' called for Sneakers::Timeout:Class"

It looks like the stdlib `Timeout` constant gets resolved to `Sneakers::Timeout`. Let me know if you insist on that name, but it will come with an overhead, probably prefixing the `::Timeout` will do, but still not ideal.

michaelklishin: @dsbonev `Sneakers::Timeout` need to be fully qualified to not conflict with `::Timeout`. I don't feel particularly strongly about the name but fully-qualifying classes with clashing names is perfectly fine and a fact of life.

michaelklishin:
Thank you!

PR #266, opened 2017-01-29
added shared connection option for worker group

mikebobrov:


michaelklishin:
@jondot @gabrieljoelc sharing a connection should be OK. Objections to this? better option names?

lib/sneakers/workergroup.rb
```diff
@@ -29,7 +29,9 @@ def run
         worker_classes = worker_classes.call
       end
 
-      @workers = worker_classes.map{|w| w.new(nil, pool) }
+      bunny_connection = config[:share_bunny_connection] ? create_bunny_connection : nil
```
gabrieljoelc: Maybe just push this into the private method?

lib/sneakers/workergroup.rb
```diff
@@ -52,5 +54,11 @@ def stop
       @stop_flag.set!
     end
 
+    def create_bunny_connection
+
```
gabrieljoelc: There's an extra space here and 62.

mikebobrov:
did the requested changes

michaelklishin:
@gabrieljoelc this generally LGTM.

lib/sneakers/workergroup.rb
```diff
@@ -52,5 +54,13 @@ def stop
       @stop_flag.set!
     end
 
+    def create_bunny_connection
+      Bunny.new(Sneakers::CONFIG[:amqp], :vhost => Sneakers::CONFIG[:vhost], :heartbeat => Sneakers::CONFIG[:heartbeat], :logger => Sneakers::logger)
+    end
+
+    def create_connection_or_nil
```
michaelklishin: I'm afraid it's not obvious what the effects of passing in a `Bunny::Session` instance vs. a `nil` are. Can you please add a comment explaining that a non-`nil` connection will be used as is, and a `nil` will ensure workers open their own?

michaelklishin: @mikebobrov sorry, I know this PR has been open for a while but before we ask @gabrieljoelc to review it one last time, can you please leave a comment as requested above? Thank you.

mikebobrov: sure. no problem. sorry for the delay. i commented out the invocation of the function. hope it's rather helpfull

michaelklishin:
@gabrieljoelc can we get your blessing to merge this? :)

gabrieljoelc:
Code looks good but is there a concern​ that existing applications using sneakers the have a larger number of consumers would have a sudden throughput drop? Here's a good [thread](https://stackoverflow.com/questions/10407760/is-there-a-performance-difference-between-pooling-connections-or-channels-in-rab) discussing this. This [answer](https://stackoverflow.com/a/24162212) talks about issues when using a cluster/load balancer and po
throughput dropping. The poster suggests using a connection pool as an alternative.

Should we consider leveraging the [`connection_pool`](https://github.com/mperham/connection_pool) gem and allowing users to configure the shared connection pool size? We could allow the default configuration to use a connection per worker by default to make it backwards compatible. This would mitigate users experiencing a possible sudden throughput drop in production.

michaelklishin:
@gabrieljoelc with this PR the default doesn't change: if `:share_bunny_connection` isn't set, each worker will open its own connection.

We have counterexamples where too many connections is a problem: #291.

Getting connection pooling right is much harder than it sounds so I'm deeply suspicious of pooling libraries by default. We can consider pooling but it's out of scope for this PR.

michaelklishin:
`connection_pool` seems to be from the author of Dalli, so it must be good and battle tested. Still, it's out of scope for this PR and we don't really know how well it works with Bunny and automatic connection recovery (in theory it should work OK).

gabrieljoelc:
@michaelklishin makes sense. I created #292 as a future alternative.

> connection_pool seems to be from the author of Dalli, so it must be good and battle tested.

Also, Sidekiq documentation suggests using [`connection_pool`](https://github.com/mperham/sidekiq/wiki/Advanced-Options#connection-pooling).

:shipit:

gabrieljoelc:
Should we add something to the README or the wiki?

michaelklishin:
@gabrieljoelc I added https://github.com/jondot/sneakers/wiki/Configuration#rabbitmq-connections a few days ago.

PR #271, opened 2017-03-27
Set properties on message to error queue

fbbergamo:
When my application tries to re-queue a error message, It lost properties data like: timestamp, app_id, message_id and headers.

This PR add the original properties in the error message. 

If anyone knows a better way to re-queue a error message, I will be very glad.


michaelklishin:
@jondot @gabrieljoelc fine with you?

dukex:
😍 Awesome!

PR #275, opened 2017-04-11
Fix shutdown

anderslemke:
I have had a problem with workers not shutting down when killed gracefully.

This seems to fix the problem for me.

The fix is basically swapping the order of `@queue.unsubscribe` and `@pool.shutdown`.

Before this, I would get a lot of R12 errors on Heroku. Probably at every restart.

Here is what Heroku says about R12:

    A process failed to exit within 30 seconds of being sent
    a SIGTERM indicating that it should stop. 
    The process is sent SIGKILL to force an exit.

After running with this in production on Heroku, I've still seen some H12. But not at every restart.

Also, when I run the spec, without swapping the order, I frequently get this error:

    2017-04-12T13:37:44Z p-55758 t-ox3doq5fw ERROR: Uncaught exception from consumer #<Bunny::Consumer:70300013132440 @channel_id=1 @queue=integration_10qfne2z2cf0btb9z90h71qj> @consumer_tag=bunny-1492004259000-703809509786>: #<RuntimeError: unable to add work while shutting down> @ /usr/local/var/rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/thread-0.1.7/lib/thread/pool.rb:252:in `block in process!'

### DIFF
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -65,7 +65,7 @@ class Sneakers::Queue
       Sneakers.logger.info "Queue: #{cancel_ok.consumer_tag} cancelled"
       @consumer = nil
     else
-      Sneakers.logger.warn "Queue: could not cancel#{@consume.inspect}"
+      Sneakers.logger.warn "Queue: could not cancel #{@consumer.inspect}"
       sleep(1)
       unsubscribe
     end
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -56,9 +56,19 @@ class Sneakers::Queue
   end
 
   def unsubscribe
+    return unless @consumer
+
     # XXX can we cancel bunny and channel too?
-    @consumer.cancel if @consumer
-    @consumer = nil
+    Sneakers.logger.info("Queue: Trying to unsubscribe on #{@consumer.inspect}")
+    cancel_ok = @consumer.cancel
+    if cancel_ok
+      Sneakers.logger.info "Queue: #{cancel_ok.consumer_tag} cancelled"
+      @consumer = nil
+    else
+      Sneakers.logger.warn "Queue: could not cancel#{@consume.inspect}"
+      sleep(1)
+      unsubscribe
+    end
   end
 
   def create_bunny_connection
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -95,10 +95,10 @@ module Sneakers
     end
 
     def stop
-      worker_trace "Stopping worker: shutting down thread pool."
-      @pool.shutdown
       worker_trace "Stopping worker: unsubscribing."
       @queue.unsubscribe
+      worker_trace "Stopping worker: shutting down thread pool."
+      @pool.shutdown
       worker_trace "Stopping worker: I'm gone."
     end
 
--- a/spec/sneakers/integration_spec.rb
+++ b/spec/sneakers/integration_spec.rb
@@ -65,7 +65,7 @@ describe "integration" do
           return
         end
       end
-      
+
       integration_log "failed test. killing off workers."
       Process.kill("TERM", pid)
       sleep 1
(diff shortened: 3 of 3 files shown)


lib/sneakers/worker.rb
```diff
@@ -92,6 +92,16 @@ def do_work(delivery_info, metadata, msg, handler)
 
         metrics.increment("work.#{self.class.name}.ended")
       end #process
+    rescue => ex
+      # If the thread pool is shutting down, it will throw a RuntimeError
+      # So, requeue the message as it is not processed.
+      handler.reject(delivery_info, metadata, msg, true)
+
+      # If is was not the thread pool shutting down, let's know about the error:
+      unless ( ex.to_s =~ /unable to add work while shutting down/i )
```
michaelklishin: It would be really nice to avoid depending on a particular exception message here. If we need to make Bunny throw a custom exception type, I'm all for it.

anderslemke: First, it's not Bunny, but Thread, throwing the error.

Second, I don't think this is a dependency. It's more of a service.
We expect Thread to throw an error, let's not make noise if it's the error we expect.

I do not see a big difference in matching on a class or a string... Both are quite fragile and depending on the underlying gem to actually throw exactly that.

Anyhow, I will investigate if messages are dropped, if passed to Sneakers by Bunny while shutting down. I have a suspicion that they actually are. That would be, sorry that I'm repeating myself, critical.

anderslemke:
I'm pretty sure messages are lost while Sneakers is shutting down.

At least I will start running this branch in production ASAP.

anderslemke:
But it would be nice to know, if there is any caveats that I haven't foreseen.

anderslemke:
Just tried it in production under heavy load, and that seemed to bring down the RabbitMQ-server.

Something is not right... I'll get back to it tomorrow...

lib/sneakers/worker.rb
anderslemke: Sorry, this was not what was needed to fix my problem.

And also, that whole "you might lose messages", was a false alarm. Sorry about that as well.

anderslemke:
I've added an integration test that will fail when the `@pool.shutdown` is called before the `@queue.unsubscribe`.

michaelklishin:
This is looking better now. I will do some functional QA later this week. @gabrieljoelc thoughts?

anderslemke:
Cool.

I'm not sure why the tests isn't passing on Travis. They are passing on my
Mac.

I'm assuming it has to do with the test of whether a process is alive or
not.

I've tried to make it work across platforms, but haven't succeeded. Any
help is appropriated.
-- 
Anders Lemke
==
Præstegårds Allé 50
2700 Brønshøj

+45 20 44 21 27
mail@anderslemke.dk
==

gabrieljoelc:
@michaelklishin yes, some smoke testing in some non-Heroku infrastructure would be good.

lib/sneakers/queue.rb
```diff
@@ -56,9 +56,19 @@ def subscribe(worker)
...
     # XXX can we cancel bunny and channel too?
-    @consumer.cancel if @consumer
-    @consumer = nil
+    Sneakers.logger.info("Queue: Trying to unsubscribe on #{@consumer.inspect}")
+    cancel_ok = @consumer.cancel
+    if cancel_ok
+      Sneakers.logger.info "Queue: #{cancel_ok.consumer_tag} cancelled"
+      @consumer = nil
+    else
+      Sneakers.logger.warn "Queue: could not cancel#{@consume.inspect}"
```
michaelklishin: there's a missing space in this message

michaelklishin:
It fell off my radar, I should be able to get to it next week.

anderslemke:
Just to be clear, this PR is really "just" about calling `@queue.unsubscribe` before `@pool.shutdown`.

On an abstract level, that seems reasonable to me, but I might be wrong.

Nevertheless, I'm running this in production (on Heroku) and it seems to have fixed my problem with workers refusing to quit when I restart our app.

michaelklishin:
@anderslemke that's a safer way to stop a consumer (well, if we have to shut the pool down explicitly at all but that's a separate conversation).

michaelklishin:
@anderslemke @gabrieljoelc since `Sneakers::Queue#subscribe` opens its own connection and channel, I would seriously consider closing those in `#unsubscribe`. In the meantime I'm OK with merging this.

PR #276, opened 2017-04-12
Ability to pass 'properties' to Bunny session

brianstorti:
With this change we will be able to define a `properties` property in the `Sneakers` configuration, that will be passed to `Bunny` when a connection is created.

```ruby
Sneakers.configure({
  amqp: 'amqp://guest:guest@localhost:5672',
  properties: {
    key: "value",
    key2: "value2"
  }
})
```

This can be useful, for example, to identify where a connection is coming from, when we have dozens of applications using a `RabbitMQ` instance.

This is something that can already be done with `Bunny`:

```ruby
Bunny.new(connection_string, { properties: { key: 'value' } })
```

So we are just leveraging that functionality when the connection is created from `Sneakers`.

These properties will be shown in the Management UI like this:

screenshot 2017-04-12 14 38 28

### DIFF
--- a/lib/sneakers/publisher.rb
+++ b/lib/sneakers/publisher.rb
@@ -34,8 +34,10 @@ module Sneakers
     end
 
     def create_bunny_connection
-      Bunny.new(@opts[:amqp], :vhost => @opts[:vhost], :heartbeat => @opts[:heartbeat], :logger => Sneakers::logger)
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you. The goal of this PR makes sense. We'll test as time permits.

PR #278, opened 2017-04-18
Exp backoff

dwkoogt:
Tried creating a exponential backoff retry handler that can be configured same way as the Maxretry handler. I have wrote about it in this [blog post](https://medium.com/@dongwookkoo/rails-sneakers-and-exponential-backoff-retry-7d92875a804b).

### DIFF
--- a/examples/exp_backoff_handler.rb
+++ b/examples/exp_backoff_handler.rb
@@ -8,7 +8,7 @@ Sneakers.configure(:handler => Sneakers::Handlers::Expbackoff,
                    :workers => 1,
                    :threads => 1,
                    :prefetch => 1,
-                   :exchange => 'sneakers-exp',
+                   :exchange => 'sneakers',
                    :exchange_options => { :type => 'topic', durable: true },
                    :routing_key => ['#', 'something'],
                    :retry_max_times => 3,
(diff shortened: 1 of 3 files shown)


michaelklishin:
@gabrieljoelc thoughts on this? It can be useful but exponential backoff also can be confusing to both humans and monitoring systems. Should we provide such implementation given that it can be provided in the user code?

gabrieljoelc:
I've contemplated the same thing for a while. I need to spend some time reviewing and see if it's extendible/configurable enough to match up with my existing use case. 😛

michaelklishin:
@gabrieljoelc did you get a chance to investigate this? :) We recently introduced a similar (pluggable) feature in the RabbitMQ Java client, so unless the exponential back-off behavior is forced on everyone, I'm a lot more open to it now. It's still damn hard to reason about it when you watch it in action, though :/

gabrieljoelc:
@michaelklishin I didn't get a chance yet and I'll be hard-pressed to find the time right now.

@dwkoogt this may be a bit much to ask, but would you be open to comparing this [other backoff handler](https://github.com/giftcardzen/la_gear/blob/develop/lib/la_gear/sneakers/exponential_backoff.rb) to discuss commonalities?

dwkoogt:
@gabrieljoelc hi. The other backoff handler uses a different approach and setup. It uses per message expiration where expiration time is a constant * (2** num_failures). I've looked into using that but didn't because the rabbitmq doc says "When setting per-message TTL expired messages can queue up behind non-expired ones until the latter are consumed or expired." (https://www.rabbitmq.com/ttl.html).

abepetrillo:
Might be worth looking at the [sneakers_handlers](https://github.com/alphasights/sneakers_handlers) as well

Issue #281, opened 2017-05-06
Dead code or what?


meetme2meat:
Looking at the sneakers code base I see [this ]( https://github.com/jondot/sneakers/blob/master/lib/sneakers/workergroup.rb#L28-L30). As I see currently the workers supplied to `serverengine` from [here](https://github.com/jondot/sneakers/blob/master/lib/sneakers/tasks.rb#L40) and [here](https://github.com/jondot/sneakers/blob/master/lib/sneakers/cli.rb#L74) are basically an `Array` object and If that is true then I believe the following condition `worker_classes.respond_to? :call` always result in `false` check.

If all the above is true I would like to propose a pull request for the removing that dead code.

michaelklishin:
That code doesn't seem to be accidental, just a fairly common idiom of supporting callables that produce final configuration values. In a dynamically typed language it's not as crazy as it may sound
and sometimes this form of lazy initialization can make a difference.

It could be specific to a particular (e.g. older) `serverengine` version, too.

PR #282, opened 2017-05-06
Remove Dead code

meetme2meat:
I'm sorry if I jump the gun. The Pull request refers to the issue highlighted on https://github.com/jondot/sneakers/issues/281. 

I ran the `test` found no failure supporting that code.


michaelklishin:
TL;DR: I don't think this is dead code.

That code doesn't seem to be accidental, just a fairly common idiom of supporting callables that produce final configuration values. In a dynamically typed language it's not as crazy as it may sound
and sometimes this form of lazy initialization can make a difference.

PR #289, opened 2017-05-24
Bumped bunny version

### LINKED ISSUE #288: Use latest version of Bunny
Bunny release v2.7.0 which include important changes on how connections are kept alive. Sneakers should be updated to support latest version. Since it are all under the hood changes it just means to bump up the gem dependency

JanStevens:
Fixes #288


michaelklishin:
Thank you!

PR #305, opened 2017-07-19
Use provided shared connection in WorkerGroup

cema-sp:
I've faced an issue with Bunny connection provided to `Sneakers.configure` and WorkerGroup.
Having the following config:

~~~ruby
connection = Bunny.new(@config)

Sneakers.configure exchange:                'ex',
                   exchange_options:        { type: :direct },
                   connection:              connection
~~~

I get a bunch of connection errors after switching from stable **2.5** release to Sneakers **master** branch.

After thorough investigation I found the root of the problem: WorkerGroup keeps creating a new connection even though I provide one in `Sneakers.configure` (and moreover it uses `:amqp` config param that I don't provide).

The solution here is using preconfigured connection in WorkerGroup and enabling it with `:share_bunny_connection` flag.

Any suggestions about PR? Should we use `:share_bunny_connection` flag or just check existence of `:connection` config param?

### DIFF
--- a/lib/sneakers/workergroup.rb
+++ b/lib/sneakers/workergroup.rb
@@ -33,9 +33,12 @@ module Sneakers
       # the queue used in the worker will create a new one
       # so if we want to have a shared bunny connection for the workers
       # we must create it here
-      bunny_connection = create_connection_or_nil
+      bunny = shared_connection_or_nil
 
-      @workers = worker_classes.map{|w| w.new(nil, pool, {connection: bunny_connection}) }
+      @workers = worker_classes.map do |worker_class|
+        worker_class.new(nil, pool, { connection: bunny })
+      end
+
       # if more than one worker this should be per worker
       # accumulate clients and consumers as well
       @workers.each do |worker|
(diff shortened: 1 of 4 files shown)


michaelklishin:
Your change makes more sense than what we currently have. It also makes the behavior in line with [what our own docs recommend](https://github.com/jondot/sneakers/wiki/Configuration). @cema-sp please amend this to ignore the `:share_bunny_connection` option and only check if `:connection` is non-nil.

michaelklishin:
@gabrieljoelc FYI.

gabrieljoelc:
I haven't gotten to releasing master as a patch version yet. We are saying this needs to go in before we do that, yes?

This test is failing:
```
integration::first#test_0001_should be possible to terminate when queue is full [/home/travis/build/jondot/sneakers/spec/sneakers/integration_spec.rb:123]:
Expected: false
Actual: true
```

michaelklishin:
@gabrieljoelc yes, it would be nice. Note that the same test fails in master.

pomnikita:
Hi, issue is relevant, let me make changes for @cema-sp , please take a look at https://github.com/jondot/sneakers/pull/322 

Thx

Issue #306, opened 2017-07-31
Provide worker class to error reporters


jdlombardozzi:
I need further context as to which worker an exception occurred. Simple solution is to provide the class to the error reporter handler. This will provide more options for the handler. In my case, I want to direct the error to a worker specific logger. I currently do a worker specific logger for each worker class, but unless I override the work method or rescue all exceptions within that method I am unable to log the error to my expected logger.

michaelklishin:
Addressed by #307.

PR #307, opened 2017-07-31
Provide worker class to error reporters #306

jdlombardozzi:
Modified worker_error method to provide current class to the handler call


michaelklishin:
@gabrieljoelc LGTM. Any objections from you?

gabrieljoelc:
Looks good. Isn't this a breaking change for someone's existing custom error reporter?

michaelklishin:
It is and it will be reflected in the version and release notes.

PR #308, opened 2017-08-02
Avoid stack level to deep when trying to serialize the exception obje…

mediazard:
Avoid stack level to deep when trying to serialize the exception object in maxretry handler

### DIFF
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -135,7 +135,7 @@ module Sneakers
             "#{log_prefix} msg=failing, retry_count=#{num_attempts}, reason=#{reason}"
           end
           data = {
-            error: reason,
+            error: reason.to_s,
             num_attempts: num_attempts,
             failed_at: Time.now.iso8601,
             payload: Base64.encode64(msg.to_s),


lib/sneakers/version.rb
```diff
@@ -1,3 +1,3 @@
 module Sneakers
-  VERSION = "2.5.0"
+  VERSION = "2.5.1"
```
michaelklishin: Never ever bump version in a pull request. It's up to the maintainers to decide when and why it should be done.

michaelklishin:
@gabrieljoelc @jondot WDYT? This loses exception stack trace information but perhaps in this particular case that's not unreasonable?

mediazard:
Can you explain better what do you mean ? 
https://github.com/reevoo/sneakers/blob/93dca2b31f6116107d00136d055304fa791bb882/lib/sneakers/handlers/maxretry.rb#L147
The reason is still exception object , so why do you think it's not possible to get backtrace info ? 
We converted to string just when we assigned to error of data object.

michaelklishin:
OK, if backtrace is included as a separate field then it's of no concern. Thanks.

robinbortlik:
@michaelklishin @gabrieljoelc @jondot do we know when this fix will be merged?
Thank you :)

gabrieljoelc:
I'll bump the gem later

PR #315, opened 2017-09-13
Use `logger.info` instead of debug for method Worker#worker_trace

yxf:


michaelklishin:
Well over 1 month without feedback => closing.

PR #316, opened 2017-09-13
Allow specify another exchange in Publisher#publish options

Varentsov:
You dont have to create separate publishers for all exchanges, now you can just write:
`Sneakers::Publisher.new.publish('Some message', exchange: 'another-exchange')`
Also you can specify exchange options too:
```
options = {
  routing_key: 'downloads',
  exchange: 'another-exchange',
  exchange_options: {
  type: :topic,
    durable: true,
    arguments: {'x-arg' => 'value' }
  }
}
Sneakers::Publisher.new.publish('Some message', options)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage/
 tmp/
 .ruby-version
 .ruby-gemset
+.idea
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,13 @@ RUN apk --update add --virtual build_deps \
 build-base ruby-dev libc-dev linux-headers \
 openssl-dev
 
-ADD . /sneakers
 WORKDIR /sneakers
 
+COPY Gemfile sneakers.gemspec ./
+COPY lib/sneakers/version.rb lib/sneakers/version.rb
+
 RUN bundle --jobs=4 --retry=3
+
+COPY . .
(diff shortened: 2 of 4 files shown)


michaelklishin:
At this point, why not use Bunny directly?

Varentsov:
Sneakers has a good interfaice, so why not to use its publisher

Varentsov:
Or you dont recommend it?

michaelklishin:
@gabrieljoelc WDYT about this? Sneakers is an opinionated project by design and such features lead some to believe that it is a "nice interface for Bunny".

michaelklishin:
@Varentsov I don't find this feature necessary in Sneakers.

Varentsov:
>"nice interface for Bunny".

I'd say it's nice interface to rabbit

michaelklishin:
I believe be5ee6bcfe9c4b9e0fc26fbc95853fbedd591b38 addressed the root cause here: obtaining a channel from a publisher wasn't possible, and once it is, it is trivial to publish to any exchange using `Bunny::Exchange#publish` and `Bunny::Channel#{fanout,topic,direct,exchange}`.

I'd do that over hacking `Sneakers::Publisher` to be more like Bunny.

PR #319, opened 2017-09-22
Use existing connection if specified in config

muffinista:
The changes in #266 break sneakers if you are specifying a `:connection` in your config. This PR should fix it. I added a test against a private method, which I normally wouldn't do, but it seemed like the expedient way to test the bug.


michaelklishin:
I believe #322 addresses the same problem, please take a look.

michaelklishin:
Duplicate of #322.

Issue #244, opened 2016-08-21
Worker instance gets re-used across different threads


kalecser:
Hi guys!

I have encountered a problem where a worker that initializes instance variables on work or work_with_params might not work properly if the thread pool has more than one thread. The observed behavior is that the instance variable gets overwritten upon execution.

Investigating further it comes down to this line:

https://github.com/jondot/sneakers/blob/aeecd535ba6fd845b9cf08dde3e9b36b039e95b4/lib/sneakers/worker.rb#L59

My question is: should we change Sneakers so a new worker instance is created per thread? The other option would be to document this _behavior_.

I am available and would be happy to help with either option ;)

michaelklishin:
I think that an instance per thread is the way to go.

kalecser:
For starters I have made a PR to fix Travis build so I can test and implement the _one worker instance per thread_ strategy. 

I have been thinking about it and maybe we should simply disallow instance variables in workers since changing the behavior to instantiate multiple worker instances might break existing code somehow.

https://github.com/jondot/sneakers/pull/247

Issue #188, opened 2015-11-22
Use Sidekiq-like error handling collection


gabrieljoelc:
*Copied from https://github.com/jondot/sneakers/pull/184#discussion_r43554582*

The `raven-ruby` gem is what Sentry uses to hook into our apps, and it includes a [shim](https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L66-L68) for Sidekiq:
```ruby
raven-ruby/lib/raven/integrations/sidekiq.rb
Sidekiq.configure_server do |config|
  config.error_handlers << Proc.new {|ex,context| Raven.capture_exception(ex, :extra => {:sidekiq => context}) }
end
```
And then Sidekiq has a [method](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/exception_handler.rb#L20) that's invoked in their processing rescues that enumerates the `config.error_handlers` collection and passes the context hash to the interface method:
```ruby
# sidekiq/lib/sidekiq/exception_handler.rb
def handle_exception(ex, ctxHash={})
  Sidekiq.error_handlers.each do |handler|
    begin
      handler.call(ex, ctxHash)
    rescue => ex
      Sidekiq.logger.error "!!! ERROR HANDLER THREW AN ERROR !!!"
      Sidekiq.logger.error ex
      Sidekiq.logger.error ex.backtrace.join("\n") unless ex.backtrace.nil?
    end
  end
end
```
Sneakers should use a similar pattern. [`Sneakers::Worker#worker_error`](https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L112-L120) should enumerate the low-level error handlers (maybe "reporters" instead to avoid ambiguity with the already existing sneakers handlers) collection instead of logging things:
```ruby
# sneakers/lib/sneakers/worker.rb
def worker_error(msg, exception)
  Sneakers.error_reporters.each do |reporter|
    reporter.call(exception, { message: msg }) # add whatever else could be help for the worker context
  end
end
```
We could include a default reporter that copied sidekiq's default error [handler](https://github.com/mperham/sidekiq/blob/7a727c61cfd3f174d32e3a231b14833b61b897fe/lib/sidekiq/exception_handler.rb#L6-L15) and would log the message like you are wanting:
```
# sneakers/lib/sneakers/exception_reporter.rb
module Sneakers
  module ExceptionReporter
    class Logger
      def call(exception, context = {})
        Sneakers.logger.warn(context) if !context.empty?
[...]

PR #323, opened 2017-10-27
Set hook to run before and after the Worker processes the message

fbbergamo:
I am using the "Raven"  (https://github.com/getsentry/raven-ruby) to send exceptions to Sentry. This library is able to store a context in a class variable and must send this context if a exception happens

The problem of using this one with Sneakers is that a worker runs in a pool of threads and It's not able clean that class before one worker and another (https://docs.sentry.io/clients/ruby/context/#rack-http-context). 

 The way that we do today is doing a overwrite in the Sneakers class and runs the method to clear the Raven context but this is bring a few complexities to the project.

My suggestion is to have a hook that runs any proc function before and after the method do_work that process the message. I also included the params on the hook to be more generic and could be used in another scenarios. 

If anyone has another suggestion or alternative I will really appreciate that.


fbbergamo:
Anyone knows what is happening with travis when runs the ruby 2.2?

aiomaster:
@fbbergamo I had the same issues with travis and I think I could fix it with this:
https://github.com/jondot/sneakers/pull/325/commits/ac2d572e73f6e938746d75ca688cc03f8e5709ee
Problem was a timing issue with the rabbitmq consumer check.

michaelklishin:
I am generally deeply suspicious of a 💯 fest from people from the same company. The change is reasonable but the problem as stated can be solved with a custom exception logger, not adding more things to the hot code path.

@gabrieljoelc do you think it's worth including?

fbbergamo:
@michaelklishin Thanks for your consideration. We are really excited to use sneakers inside our organization and we are looking forward to helping improve the Lib. 

My first approach to solve this problem was using the exception report. But this doesn't solve the problem because the error context it's inside the `work` method. 

I agree with you that is possible to create a custom exception logger because we have the same information that the `work_with_params` has. But this exception logger could be generic for all the projects inside a system and the extra_context its the way to go to include extra params for a specified work and its possible to do something like this:

```
def work(message)
  id = get_something_from_db()
  Raven.extra_context(object_id: id)
  do_something()
end
```

This hooks may also help to include specific logs and metrics to run before and after the `work` as well

gabrieljoelc:
Thanks for contributing.

Have you looked at all of these existing interfaces for logging, metrics, and errors?
* [`Sneakers::Concerns::Logging`](https://github.com/jondot/sneakers/blob/master/lib/sneakers/concerns/logging.rb) - abstraction for the logger ([docs](https://github.com/jondot/sneakers/wiki/How-To:-Do-Log-Processing))
* [`Sneakers::Concerns::Metrics`](https://github.com/jondot/sneakers/blob/master/lib/sneakers/concerns/metrics.rb) - for capturing basic job counters and timing ([docs](https://github.com/jondot/sneakers/wiki/Metrics))
* [`Sneakers::ErrorReporter`](https://github.com/jondot/sneakers/blob/master/lib/sneakers/error_reporter.rb) - for passing errors to error reporters (i.e. Raven/Sentry). I need to write the docs for this but here's the original issue description that could be helpful: https://github.com/jondot/sneakers/issues/188#issue-118279624.
* Sneakers handlers - interface for handling message semantics ([docs](https://github.com/jondot/sneakers/wiki/Message-handling-semantics))

Is there a design you do that can take advantage of these existing interfaces?

fbbergamo:
Hello @gabriel
I am already taking advantage of "Sneakers::ErrorReporter" to format the error before to send to our error tracker, but I can't make sure that the singleton will be cleaned inside this "proc" because the context has been defined inside the work method.
 
The other way that I can do without this hook is to include a method in the handler that will be called before runs the work and after, What do you think?

This implementation its generic and can be also be used in other scenarios and It's not covered in any of the previous structures.

lib/sneakers/worker.rb
```diff
@@ -120,6 +124,25 @@ def worker_trace(msg)
       logger.debug(log_msg(msg))
     end
 
+    def before_work(delivery_info, metadata, msg)
+      wbefore = Sneakers::CONFIG[:hooks][:before_work]
```
dukex: Hey @fbbergamo why not rename this `wbefore` and the `wafter` to `hook`, wbefore and wafter sounds weird in the code

dukex:
@gabrieljoelc @michaelklishin and @fbbergamo Although this pull request trying to resolve a 'individual' issue, and maybe we have better ways to handle with this issues, I think this hooks could be a nice feature to have in sneakers. That give me a better and clearer way to handle with prerequisites and post requisites of my workers.

michaelklishin:
@dukex so what are you saying, this PR is worth merging as is (well, perhaps with some renaming) in your opinion or we need a more "generalised" approach?

@gabrieljoelc same question for you ^^^

gabrieljoelc:
@michaelklishin I reread https://github.com/jondot/sneakers/pull/323#issuecomment-341689801 and here are some thoughts:

**The use case:**
As a application developer
So that I can explicitly send application concerns (i.e. `object_id: id`) to my error report (i.e. `Raven. extra_context`)
I want a way to inject the application context into the error reporter

**The implementation:**
This use case is best solved with a custom error reporter:
```ruby
class ApplicationConcernedWorker
  // exposes necessary context to infrastructure-concerned error reporter
  attr_reader :id

  def work(message)
    @id = get_something_from_db()
    do_something()
  end
end

// infrastructure concerns
class RavenObjectIdReporter
  def call(exception, worker, context_hash)
    Raven.extra_context(object_id: worker.id)
  end
end

# /config/initializers/sneakers.rb
// you can add as many cross-cutting concerns as you like
// because it iterates through all of them!
// (see https://github.com/jondot/sneakers/blob/master/lib/sneakers/error_reporter.rb#L16-L24)
Sneakers.error_reporters.concat([RavenObjectIdReporter.new,
  OtherInfrastructureConcernReporter.new,
  ReportersForDays.new])

Sneakers.error_reporters << proc do |ex, worker, ctx_hash|
  AllTheReporting.notify(ex, ctx_hash)
end
```

From https://github.com/jondot/sneakers/pull/323#issuecomment-344253356:
> That give me a better and clearer way to handle with prerequisites and post requisites of my workers.

@dukex if you have additional use cases, can you list them please?

dukex:
@michaelklishin for me, this PR is worth merging as is.

@gabrieljoelc I had a question about your implementation example, it's a good idea set an instance variable to the workers, as you do with `@id`? I had many issues about concurrent workers running with this approach, the instances variables from one message was override the others, my guess is because the [worker object are shared](https://github.com/jondot/sneakers/blob/master/lib/sneakers/queue.rb#L50), make it sense? 

About your question for me, I can imagine some cases, for example, check/open/close the database connection before or after a work execution or parse the message payload given the message content_type (the @fbbergamo implementation don't allow it :disappointed:). 

Take a look this example I extract from a company I'm working https://gist.github.com/dukex/6852b1ae72999dc7c88afcab7498fc57, I know many things can be improved in this code, but this kind of things(workers prerequisites and post requisites) it's not so clear or is not so easy to do with the current sneakers API.

fbbergamo:
@gabrieljoelc 

More problems related to instance variables inside the worker 
https://github.com/jondot/sneakers/issues/244

I think that the problem is that the threads will run inside this block and they share the same work instance
https://github.com/jondot/sneakers/blob/aeecd535ba6fd845b9cf08dde3e9b36b039e95b4/lib/sneakers/worker.rb#L48

gabrieljoelc:
@dukex @fbbergamo thanks for bring up those points. You are right. There are some thread-safety issues.

Should we treat that as a bug to fix instead of working around it with the hooks?

gabrieljoelc:
I think we should revisit this PR after we fix the thread-safety issue in separate PR.

andrykonchin:
@michaelklishin @gabrieljoelc 

What would you say about adding middleware mechanism? It's a common used and more general approach in comparison to before/after hooks.

Moreover some `sneakers` internals could be refactored to built in middleware e.g. error reporting and metrics gathering.

It should have a pretty straightforward implementation and if you are for this feature I will be happy to work on it.

gabrieljoelc:
@andrykonchin I agree with you. We'd grateful for your help!

andrykonchin:
@gabrieljoelc PR is ready. Could you review please?

I am not familiar with `rr` so I will be grateful for any advise to improve specs.

PR #324, opened 2017-10-30
Refactor New Relic usage

adamors:
Metrics were recorded through a deprecated API, added support for the newer one.

As you can see here, the `get_stats` method was deprecated and the suggested change is 
> If you had chained `get_stats` with `record_data_point`, use:
`NewRelic::Agent.record_metric`

https://docs.newrelic.com/docs/agents/ruby-agent/installation/upgrade-ruby-agent-versions


michaelklishin:
Thank you.

PR #325, opened 2017-10-30
refactor publisher connection logic

aiomaster:
* separate connection stuff from publish method to better reuse Sneakers::Publisher for own derived classes
* fix warning: instance variable @bunny not initialized when connected? is called


michaelklishin:
Thanks a lot for ac2d572e73f6e938746d75ca688cc03f8e5709ee.

PR #331, opened 2017-11-15
Casually start defining classes based on jobs defining perform in Rails

### LINKED ISSUE #330: Does not handle processing queues in ActiveJob
I wanted to link together this gem and [this stale Rails/Rails issue](https://github.com/rails/rails/issues/29478) about Sneakers not handling reading the specified Queue in it's adapter.

Could we update the docs [[1](https://github.com/jondot/sneakers/wiki/How-To:-Rails-Background-Jobs-with-ActiveJob), [2](https://github.com/jondot/sneakers/wiki/How-To:-Rails-Background-Jobs)] with information on how to process a specific queue?

Perhaps as a solution you would accept a PR that enhances the Rake task with the queue's to process?

BookOfGreg:
Fixes #330 
Also attempt at rails/rails#29478

This PR defines one worker per Job that defines Perform, and sets it's queue to whatever queue_name resolves to after processing.

### DIFF
--- a/lib/sneakers/tasks.rb
+++ b/lib/sneakers/tasks.rb
@@ -37,6 +37,20 @@ EOF
       exit(1)
     end
     opts = (!ENV['WORKER_COUNT'].nil? ? {:workers => ENV['WORKER_COUNT'].to_i} : {})
+    if defined?(::Rails) && defined?(::ActiveJob::Base)
+      workers -= [ActiveJob::QueueAdapters::SneakersAdapter::JobWrapper]
+      jobs = ::ActiveJob::Base.descendants
+      not_abstract_jobs = jobs.select { |j| j.instance_methods.include?(:perform) }
+      workers += not_abstract_jobs.map do |job|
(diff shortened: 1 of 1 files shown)


BookOfGreg:
This solution has a flaw, if a worker crashes, it's fake class is no longer defined and can't be relaunched.

gabrieljoelc:
Thanks for trying to fix this. I've been on the fence about Rail's attempt to abstract background jobs for this reason. The API's are often the same (i.e. `perform` and `work` methods) but the threading/process behavior must vary as folks explore different implementations.

BookOfGreg:
I think I'll close this for now, I wouldn't want anyone to actually use it, this experimental code is pretty rough and has flaws, its more for education.

The reason for my attempt is that the interface for  an ActiveJob is so close to a worker implementation here that I feel it should be a refactor of the adapter to get the queue behaviour I wanted but I couldn't unravel the module loader thing in activejob itself.

I'll probably move the issue onto rails and re-open an issue there, see if I can make progress on this.

PR #333, opened 2017-11-30
Worker group ignores connection option

raskhadafi:
Started to use the connection option and realize that allways localhost was the target instead of another  host.
The tests are runnig but unfortunately I did not managed to create a test for it.

I hope the fix will be merge otherwise give me a short feedback.
Thanks.

### DIFF
--- a/lib/sneakers/configuration.rb
+++ b/lib/sneakers/configuration.rb
@@ -79,6 +79,7 @@ module Sneakers
           hash.delete(k)
           @hash.delete(k)
         end
+        hash.merge!(share_bunny_connection: true) # Share the connection when set.
       end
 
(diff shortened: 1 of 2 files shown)


lib/sneakers/configuration.rb
```diff
@@ -79,6 +79,7 @@ def merge!(hash)
           hash.delete(k)
           @hash.delete(k)
         end
+        hash.merge!(share_bunny_connection: true) # Share the connection when set.
```
michaelklishin: This forces `share_bunny_connection`. Is the intent to simply change the default?

michaelklishin:
Closing due to inactivity.

PR #345, opened 2018-03-01
Force eager load

jkgit:


Spaceghost:
I didn't noticed the closed state before tossing out a review, oops.

PR #346, opened 2018-03-21
Fix Error Reporters Usage Description

Tensho:
According to the definition [here](https://github.com/jondot/sneakers/blob/master/lib/sneakers/error_reporter.rb#L5) and default reporter call [here](https://github.com/jondot/sneakers/blob/master/lib/sneakers/error_reporter.rb#L18).

Also reflected error reporters at "Configuration" wiki page, new ["Error Reporters"](https://github.com/jondot/sneakers/wiki/Configuration#error-reporter) section.

### DIFF
--- a/lib/sneakers.rb
+++ b/lib/sneakers.rb
@@ -79,7 +79,7 @@ module Sneakers
 
   # Register a proc to handle any error which occurs within the Sneakers process.
   #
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

PR #347, opened 2018-03-21
Isolate the code that is processed inside the pool process into a method

fbbergamo:
This will isolate the method that is run inside the `pool.process`. This will allow us to implement any code around this block that needs to be threads safe.

This will not break any interface.


michaelklishin:
Thank you!

PR #351, opened 2018-04-24
Declare Rake as Runtime Dependency

Tensho:
Sneakers runs primary within rake task `sneakers:run`, so it's fair to have `rake` gem as runtime dependency along with `thor`. This change will allow running `sneakers` in a bare Ruby application without `gem "rake"` in `Gemfile`.


michaelklishin:
@Tensho fair enough. Please rebase this against master?

PR #352, opened 2018-05-03
Daemonize the sneakers started using rake task

meetme2meat:
Currently, we use screen and nohup options to push the start the task in background mode. Hopefully after this fixes we would no longer have to do that.

Once accepted that the following options (I will) need to be updated on wiki page over [here](https://github.com/jondot/sneakers/wiki/How-To:-Rails-Background-Jobs)


meetme2meat:
I release sneaker has an option to daemonize. Closing this PR.

PR #353, opened 2018-05-28
change unnecessary conditions

kamalhosny:
.blank? method checks for all falsy expressions including nil? empty?
that will make it more readable


kamalhosny:
> An object is blank if it’s false, empty, or a whitespace string. For example, false, ”, ‘ ’, nil, [], and {} are all blank.
this is the documentation link : 
[https://apidock.com/rails/Object/blank%3F](url)
so nil.blank? is true : 
`2.4.0 :001 > nil.blank?
 => true `
I guess it's readable even in normal English @michaelklishin

PR #354, opened 2018-06-03
BREAKING CHANGE: remove enforced Worker#work timeout

### LINKED ISSUE #343: Alternative timeout mechanism worker operations (the work method)
Hi,

Could you please remove the use of `Timeout::timeout` from line 55 of `worker.rb`? This function is [inherently unsafe](http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html) and will potentially 'randomly' kill workers. 

At the very least, would it be possible to default `:timeout_job_after` to 0 so that timeouts are switched off by default? Then people can decide for themselves whether they want to run the risks of `Timeout::timeout` in their code.

I've been hit by this a few times now and finally understood what's going on. Given that the timeout exceptions spring up in random places in my worker code (in fact, deep in some random Rails library) it took a long time to understand that it was actually Sneakers breaking stuff and not my own code. The ability to time out is fairly well hidden in the documentation, so I only found it when I specifically searched for timeout in connection to Sneakers.

Many thanks,

Steffen

michaelklishin:
See #343 for the background. Worker implementations now must handle
timeouts in a way they see fit. This is unfortunate but there is no
way to enforce timeouts in a safe way without at least some Worker implementation
cooperation and currently there is no decision on what that should look like.

We previously increased the timeout to 10 minutes. This is about as good
as not having any, so let's rip off the bandaid.

Closes #343.

### DIFF
--- a/lib/sneakers/configuration.rb
+++ b/lib/sneakers/configuration.rb
@@ -37,7 +37,6 @@ module Sneakers
       :amqp_heartbeat     => 30,
 
       # workers
-      :timeout_job_after  => 600,
       :prefetch           => 10,
       :threads            => 10,
       :share_threads      => false,
--- a/lib/sneakers/errors.rb
+++ b/lib/sneakers/errors.rb
@@ -1,5 +1,2 @@
-require 'timeout'
-
 module Sneakers
-  class WorkerTimeout < Timeout::Error; end
 end
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -5,7 +5,7 @@ module Sneakers
   module Handlers
     #
     # Maxretry uses dead letter policies on Rabbitmq to requeue and retry
-    # messages after failure (rejections, errors and timeouts). When the maximum
+    # messages after failure (rejections and errors). When the maximum
     # number of retries is reached it will put the message on an error queue.
     # This handler will only retry at the queue level. To accomplish that, the
     # setup is a bit complex.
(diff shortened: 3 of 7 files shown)


michaelklishin:
@gabrieljoelc you haven't participated in the heated discussion in #343. Do you have any objections or alternative solution ideas for this? Thank you.

gabrieljoelc:
I somehow missed that entire conversation. I'll take a look later today

gabrieljoelc:
Read through #343 . Is the main desire to either 1. control it manually through custom timeout code and/or 2. through the infrastructure killing processing that run too long?

Should we add a recommendation to add timeout code in the readme?

Tensho:
@gabrieljoelc Yep, that's exactly the plan. We didn't have a talk about "infrastructure killing processing", but I think it's a good point to mention in the documentation.

michaelklishin:
I don't know if this deserves a major version bump, although we probably can. My vote is for 2.10.

Tensho:
My sentiments the same. Many big Ruby libraries do the same, for example, `mail` gem.

michaelklishin:
@Tensho sorry, do what specifically?

Tensho:
I mean some wide used gems do not adhere semver strictly and may release breaking changes to public API with note in CHANGELOG.

michaelklishin:
Bumping major version every time a breaking change, regardless of the scope, is introduced, is pretty radical and not too many projects do that in practice.

We can bump it since Sneakers hasn't done that in a long time. Whether this change warrants a major version bump is a matter of opinion.

Tensho:
I vote for `2.10`

PR #364, opened 2018-07-24
Introduce middlewares

andrykonchin:
Related issue https://github.com/jondot/sneakers/pull/323

It allows to use custom hooks/middlewares while job is processed.
The interface is similar to Rails:

```ruby
Sneakers.middleware.use(<class>, arguments)
Sneakers.middleware.delete(<class>)
```

Rails documentation - https://guides.rubyonrails.org/rails_on_rack.html#configuring-middleware-stack

Middleware should have following interface:

```ruby
class Middleware
  def initialize(app, <arguments>)
    ...
  end

  def call(message, delivery_info, metadata, handler)
    ...
  end
end
```

`#call` method should
- call next middleware passed into `#initialize` as `app` argument
- pass `message`, `delivery_info`, `metadata`, `handler` as arguments and
- modify and return result value.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,20 +9,16 @@ RUN apk --update add --virtual build_deps \
                                linux-headers \
                                openssl-dev
 
-RUN mkdir /myapp
-
 WORKDIR /sneakers
 
-COPY lib/sneakers/version.rb /sneakers/lib/sneakers/version.rb
+COPY lib/sneakers/version.rb lib/sneakers/version.rb
 
-COPY sneakers.gemspec /sneakers/sneakers.gemspec
+COPY sneakers.gemspec .
 
-COPY Gemfile /sneakers/Gemfile
-
-COPY Gemfile.lock /sneakers/Gemfile.lock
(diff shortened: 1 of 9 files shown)


michaelklishin:
@gabrieljoelc what do you think about the general approach?

gabrieljoelc:
@andrykonchin thanks for doing this. Looks pretty good. Can you create a sample middleware in the samples to see it's use?

andrykonchin:
@gabrieljoelc Done

gabrieljoelc:
@michaelklishin you good with this?

dukex:
Nice job @andrykonchin!

gabrieljoelc:
@michaelklishin have you had a chance to review this?

michaelklishin:
I think this should be seriously considered after the next release (which is long overdue).

alipio:
Well done @andrykonchin! 💯

fbbergamo:
Nice job @andrykonchin 
@michaelklishin  Is there anything that may help to get this merged?

PR #366, opened 2018-08-15
Add push message with ruby in README

darren987469:
I just started to use sneakers and rabbitmq. Following the quick start, it takes me a while to figure out how to push message to the logs queue. I added some code that I used to push message with ruby. That would help beginner feel less frustrated and start quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 
 ```
       __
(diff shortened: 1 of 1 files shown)


gabrieljoelc:
Thanks for you contribution!

PR #372, opened 2018-10-18
Loosen Bunny dependency to minor version

olivierlacan:
Sneakers is currently locking us into a version of Bunny we don't want to be using because the version dependency is tied to the patch-level. Is there a reason for that? 

Currently we're locking people into patch versions of Bunny when the project clearly seems to be using SemVer and shouldn't have API breaking changes in minor versions. So this change should make sense.

See: https://github.com/ruby-amqp/bunny/blob/master/ChangeLog.md

### DIFF
--- a/sneakers.gemspec
+++ b/sneakers.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
 
   gem.add_dependency 'serverengine', '~> 2.0.5'
-  gem.add_dependency 'bunny', '~> 2.11.0'
+  gem.add_dependency 'bunny', '~> 2.12'
   gem.add_dependency 'concurrent-ruby', '~> 1.0'
   gem.add_dependency 'thor'
   gem.add_dependency 'rake'


olivierlacan:
@michaelklishin could you review and/or merge this? Merging it would allow people to benefit from the queue declaration timeout fix in Bunny 2.12. 

See: 
- https://github.com/ruby-amqp/bunny/pull/566
- https://github.com/ruby-amqp/bunny/issues/558

Thanks 🤗

lucaong:
Any update on this? It would be really good to be able to use the most recent version of Bunny

olivierlacan:
@gabrieljoelc Any way I can help you cut a release for this?

lucaong:
I would also be happy to help, if anything is blocking the creation of a new release including this change.

gabrieljoelc:
@olivierlacan @lucaong thanks for the offer! i started going through the commits that are different between the last release and master but didn't have time to finish and build a changelog entry.

If you are able to help with this, that'd be great. Can you go through those commits and create a PR for the [changelog](https://github.com/jondot/sneakers/blob/master/ChangeLog.md) for the new release? If you are able to, can you add any necessary updates to the readme as well please?

Looks like we are behind on [release tags/"releases"](https://github.com/jondot/sneakers/releases) and [published gem versions](https://rubygems.org/gems/sneakers) for 2.8.0 and 2.10.0. If you guys can focus on the changelog and start discussion about the semver number the next release can be, I can try to backfill the other two versions this weekend.

Thanks again for your generous offers!

lucaong:
@gabrieljoelc and @olivierlacan, I just compiled a list of changes for the changelog and sent a PR: https://github.com/jondot/sneakers/pull/378

Feel free to comment on it and let me know if anything is missing

gabrieljoelc:
I believe this was the much wanted feature for the the [`v2.11.0` release](https://github.com/jondot/sneakers/releases/tag/v2.11.0), so cc'ing https://github.com/jondot/sneakers/pull/378#issuecomment-455807025 here as well.

PR #376, opened 2018-12-08
Allow handlers to configure queue

jbrady42:
Adds a hook on handler class to allow configuration of options before queue creation and handler instantiation.

This allow Maxretry handler to auto-configure the required dead letter exchange on the worker queue.

### DIFF
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -82,6 +82,15 @@ module Sneakers
 
       end
 
+      def self.configure_queue name, opts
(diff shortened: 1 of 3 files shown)


lib/sneakers/handlers/maxretry.rb
```diff
@@ -82,6 +82,15 @@ def initialize(channel, queue, opts)
 
       end
 
+      def self.configure_queue name, opts
```
michaelklishin: I don't like asking for this but can we please use the existing code style for method definitions?

michaelklishin:
This looks reasonable and backwards compatible at first glace 👍

michaelklishin:
Thank you!

PR #387, opened 2019-03-07
support multiple exchanges for queue

stevereinke:
Hi there, 

in our project we tried to use two delivery strategies for one queue. We want to connect our queue to a exchange of type direct and to another exchange of type fanout. So we changed the subscrive method of the queue object to support multiple exchanges. Please have a look at the code and merge it if it's fine. ;-)

Kind regards, 
Steve

### DIFF
--- a/lib/sneakers/queue.rb
+++ b/lib/sneakers/queue.rb
@@ -1,11 +1,14 @@
-
 class Sneakers::Queue
-  attr_reader :name, :opts, :exchange, :channel
+  attr_reader :name, :opts, :exchanges, :channel
 
   def initialize(name, opts)
     @name = name
     @opts = opts
     @handler_klass = Sneakers::CONFIG[:handler]
+  end
+
+  def exchange
+    exchanges.first if exchanges
   end
 
   #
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you for your time and interest in contributing!

The number of small features that control the topology is getting out of hand. Sneakers cannot accommodate every possible scenario and adding more and more features in this are is not sustainable.

I recommend setting it up with Bunny instead (Bunny connection and channel should be exposed via the API). I'd prefer to see more interest in this feature before we merge it.

michaelklishin:
@Tensho havee you encounters similar scenarios? Is there anything that would make it hard[er] to use Bunny directly to set up the topology? I'd rather work on fixing that then adding more and more topology definition features.

Tensho:
@michaelklishin, my sentiments the same – I used to consider `sneakers` in a [worker queue](https://www.rabbitmq.com/tutorials/tutorial-two-ruby.html) scenario mostly. Despite I don't recommend to go this way either, it's possible to control more complex topologies with the **different `Sneakers::Worker` definitions**. TBH, I have never seen the need to bind multiple exchanges to one queue, only vice versa bindings.

PR #388, opened 2019-03-19
[WIP] Add option for making consumer exclusive

anderslemke:
- [ ] Documentation

In some situations, the execution order of messages is crucial.
In such situations, RabbitMQ can ensure that there are no more than one consumers on a queue, [with the exclusive option](https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume).


michaelklishin:
This does not make [the consumer exclusive](https://www.rabbitmq.com/consumers.html#exclusivity), it makes the queue exclusive. Is that what you are really after?

anderslemke:
@michaelklishin I might confuse some concepts here.

There is an `exclusive` flag in this gem already:
https://github.com/jondot/sneakers/blob/7c059d1490e4dde43f98b801c38e4e0df7ac458c/lib/sneakers/configuration.rb#L20
As far as I understand, this controls the queue: https://www.rabbitmq.com/queues.html#exclusive-queues . This is what I would call "exclusive queue".

I want this PR to control the `exclusive`-flag in https://www.rabbitmq.com/amqp-0-9-1-reference.html#basic.consume . This is what I refer to when I say "exclusive consumer".

I will beef up the PR-description to talk about the motivation for this.

anderslemke:
@michaelklishin I've tested this PR locally, and I get the following results.

When I set `exclusive` to `false`, I get consumers with this config:

Screenshot 2019-03-20 at 10 13 50

When I set `exclusive` to `true`, I get a consumer with this config: 

Screenshot 2019-03-20 at 10 13 32

This is actually what I want to achieve. But I'm not sure I use the terms correct, though.

PR #390, opened 2019-04-16
Allow re-raise configured exceptions to worker

Curter29:
If we need process some exceptions inside worker
before
```
rescue StandardError, ScriptError => ex
```

### DIFF
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -72,6 +72,8 @@ module Sneakers
           end
           res = block_to_call.call(deserialized_msg, delivery_info, metadata, handler)
         end
+      rescue *opts[:raise_exceptions] => ex
+        raise
       rescue StandardError, ScriptError => ex
         res = :error
         error = ex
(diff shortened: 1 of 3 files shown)


michaelklishin:
Please handle exceptions in your own code. Sneakers handles exceptions as a last resort. RabbitMQ consumers are generally expected to deal with their own exceptions.

@Tensho OK with you if I close this?

Tensho:
@michaelklishin  Sure, totally agree with your point.

lib/sneakers/worker.rb
```diff
@@ -72,6 +72,8 @@ def process_work(delivery_info, metadata, msg, handler)
           end
           res = block_to_call.call(deserialized_msg, delivery_info, metadata, handler)
         end
+      rescue *opts[:raise_exceptions] => ex
+        raise
```
anderslemke: Wouldn't it make sense to re-raise existing exception?

```suggestion
        raise ex
```

anderslemke:
@Curter29 I would like to understand why you would want to have any of `StandardError` or `ScriptError` bubble up to your app?

Also, raising exceptions at this point in the code, would make it ambiguous what happens to messages that are fetched from Rabbit. I would say, that they should be handled regardless of any exceptions.

PR #391, opened 2019-05-14
CI: Use 2.6.3, 2.5.5, 2.4.6

olleolleolle:
This PR only updates the CI matrix to the latest generally available Ruby versions.

Source for version numbers: https://github.com/rvm/rvm/blob/master/config/known

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,13 @@ before_script:
   - until sudo lsof -i:5672; do echo "Waiting for RabbitMQ to start..."; sleep 1; done
 rvm:
   - ruby-head
-  - "2.6.1"
-  - "2.5.3"
-  - "2.4.5"
+  - "2.6.3"
+  - "2.5.5"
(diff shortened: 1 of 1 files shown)


olleolleolle:
This fails because the systemctl start of the rabbitmq-server fails. 

Perhaps there are other ways of install now?

olleolleolle:
Services: rabbitmq, according to the Travis docs.

olleolleolle:
https://docs.travis-ci.com/user/database-setup/#rabbitmq

olleolleolle:
Hmm. Hm. Xenial dist has perhaps changed?

olleolleolle:
That fixed it.

michaelklishin:
@olleolleolle we should do [what Bunny does](https://github.com/ruby-amqp/bunny/blob/master/bin/ci/install_on_debian.sh), that is, install RabbitMQ manually without relying on an outdated version available on Travis.

PR #394, opened 2019-06-19
Fix broken build

pnikonowicz:
It appears as if the tests are expecting rabbit to be running on port 15672 instead of 5672. 

rabbit is running on 5672, point the tests to that port

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,6 @@ services:
     image: rabbitmq:management-alpine
     ports:
       - "5672:5672"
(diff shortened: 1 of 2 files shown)


pnikonowicz:
15672 is not the correct port. 
Perhaps the tests should not be looking for 15672 but instead look for 5672

michaelklishin:
This is wrong. The line in question sets up an HTTP API client, and port `5672` is for AMQP 0-9-1 and AMQP 1.0 clients.

PR #402, opened 2019-07-26
Add configuration options for consumer subscribing to a queue

hedacharon:
When a worker subscribes to a queue only two configuration options were used `:block` (hardcoded) and `:ack` (configurable using `:ack` option). Setting other consumer configuration options such as [consumer priorities](http://rubybunny.info/articles/queues.html#consumer_priorities) or [exclusive consumers](http://rubybunny.info/articles/queues.html#exclusive_consumers) was not possible. 

Creating new `consumer_options` configuration options block resolves that and all configured consumer options are directly passed to `queue.subscribe` (in more general way than [pull request for exclusive consumers](https://github.com/jondot/sneakers/pull/388)) while original defaults are kept as new defaults. The approach is similar to `exchange_options` and `queue_options` which are already in codebase. Old `:ack` configuration option becomes deprecated, but is mapped via deprecated options mapping.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,6 @@ source 'https://rubygems.org'
 
 gem 'ruby-prof', platforms: [:ruby_22, :ruby_23, :ruby_24]
 
+gem 'rake', '>= 12.3', '< 14.0'
+
 gemspec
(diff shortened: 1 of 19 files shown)


michaelklishin:

PR #411, opened 2019-10-17
Zeitwerk mode

### LINKED ISSUE #396: rake sneakers:run fails on Rails 6, new default code loader related issue.
When running `rake sneakers:run` on Rails 5 application without specify `WORKERS` environment variable, sneakers will look for the `Sneakers::Worker::Classes`, which is appended when include `Sneakers::Worker` in class.

This works on Rails 5 app. However, Rails 6 changed its default code loader to zeitwerk, running sneakers rake task won't load workers class.

sneakers-run

This can be fixed by adding
```
::Zeitwerk::Loader.eager_load_all if defined?(::Zeitwerk)
```
in `lib/sneakers/tasks.rb` file.

Any thought? Would a PR be welcomed?

vladimirtemnikov:
Resolves #396 

Rails 6.0 provides Zeitwerk mode and uses different eager load logic: it calls `::Zeitwerk::Loader.eager_load_all` in `Rails::Application::Finisher`. So we need to call `eager_load_all` manually.

### DIFF
--- a/lib/sneakers/tasks.rb
+++ b/lib/sneakers/tasks.rb
@@ -10,7 +10,11 @@ namespace :sneakers do
     Rake::Task['environment'].invoke
 
     if defined?(::Rails)
-      ::Rails.application.eager_load!
+      if defined?(::Zeitwerk)
+        ::Zeitwerk::Loader.eager_load_all
+      else
+        ::Rails.application.eager_load!
+      end
     end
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

ricardovj:
When can we expect this to be released in a new gem version? Looks like the latest 2.12.0 doesn't include this fix.

PR #414, opened 2019-11-06
passing worker to the middleware call

redwrx:
a minor change to pass the worker object to the middleware call
https://github.com/jondot/sneakers/issues/415

### DIFF
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -56,6 +56,7 @@ module Sneakers
       begin
         metrics.increment("work.#{self.class.name}.started")
         metrics.timing("work.#{self.class.name}.time") do
+          metadata[:headers][:worker_name] = self.class.name if metadata && metadata[:headers]
           deserialized_msg = ContentType.deserialize(msg, @content_type || metadata && metadata[:content_type])
 
           app = -> (deserialized_msg, delivery_info, metadata, handler) do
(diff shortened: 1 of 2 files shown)


mikker:
We'd love to have this merged as it would make the Sneakers jobs look much better in Elastic APM (see https://github.com/elastic/apm-agent-ruby/pull/676#issue-360412848) ➕!

jondot:
Thanks, looks like a great idea. would this break compatibility with existing middleware? (if so, is there a way to make it backwards compatible?)

mikker:
Seems to me this would be breaking as the position of all arguments get shifted. Without knowing much about the rest of the codebase, one way to not be breaking would be to add `self` to one of the currently passed arguments, eg `metadata[:worker] = self`.

jondot:
Interesting. I would also accept having two different signatures, one with this and one without (older one).
Would you be willing to submit a PR or augment this one to be backward compatible? if so I'd be happy to merge of course

redwrx:
> Interesting. I would also accept having two different signatures, one with this and one without (older one).
> Would you be willing to submit a PR or augment this one to be backward compatible? if so I'd be happy to merge of course

I agree and will update this PR to follow @mikker approach

redwrx:
@jondot  since `metadata` was a  bunny MessageProperties object, I passed the `worker_name` to the `headers` hash inside metadata.

redwrx:
> Interesting. I would also accept having two different signatures, one with this and one without (older one).
> Would you be willing to submit a PR or augment this one to be backward compatible? if so I'd be happy to merge of course
@jondot 
I pushed the code that uses metadata to pass around the worker name. It passes all tests for all ruby versions, except Ruby-head. Seems like there is a bug with minitest that prevents the tests to run successfully 
`ArgumentError: wrong number of arguments (given 2, expected 1)`

https://github.com/seattlerb/minitest/issues/689

PR #421, opened 2019-12-19
Use ERB in yaml in order to use env vars to control worker count/etc

rsouza-mdlive:
We rely on the WorkerGroup yml file to divide our workers, but we lacked control over the worker count per group.

This PR is to enable use ENV vars (via erb interpolation)  inside the yml file.


michaelklishin:
Fair enough.

PR #423, opened 2020-01-09
Change publisher message to use debug log level

abezzub:
Sneakers uses inconsistent log level for messages produced by worker and publisher: worker uses `debug`, but publisher uses `info`. I think it makes sense for them to use the same log level, so it will be easy to configure logger to the right log level.

Worker:
https://github.com/jondot/sneakers/blob/master/lib/sneakers/worker.rb#L123


abezzub:
Test failure appears to be unrelated to the PR.

PR #427, opened 2020-02-18
Preserve queue arguments when MaxRetry handler is used

sharshenov:
`2.12.0+` versions of Sneakers have MaxRetry handler incompatible with custom queue arguments.

```ruby
class AwesomeWorker
  include Sneakers::Worker

  from_queue 'awesome',
             handler: Sneakers::Handlers::Maxretry,
             arguments: {
               'x-max-priority' => 10 # <-- this argument is ignored
             }

  def work(_)
    ack!
  end
end
```


sharshenov:
Tests are failing for rub-head like in master because of `metric_fu-4.12.0`

mwannewitz:
What is the state of this PR? We would really like to make use of the suggested improvement.

michaelklishin:

PR #428, opened 2020-02-19
TTL for the error queue

AlexanderPower:
Added ability to automatically delete messages from the error queue after message TTL has expired.


michaelklishin:

Issue #426, opened 2020-02-07
Unlock Rack gem version


morgoth:
I can see in https://github.com/jondot/sneakers/commit/591d2131b3b9d099e5b4401a0921335886ed9a2f that rack gem version was locked.
There is already 13.x of rack gem which is not allowed by this restriction.
Is there any reason to lock it or can it be relaxed?

rpbaltazar:
Related to #422

bumi:
The commit closing #422  seems to miss the development dependency: https://github.com/jondot/sneakers/commit/415f283f7e634601528105323d0dac9663c5786d

`gem.add_development_dependency 'rake', '~> 12.3'`
why is it a development and runtime dependency?

rpbaltazar:
> The commit closing #422 seems to miss the development dependency: [415f283](https://github.com/jondot/sneakers/commit/415f283f7e634601528105323d0dac9663c5786d)
> 
> `gem.add_development_dependency 'rake', '~> 12.3'`
> why is it a development and runtime dependency?

That's very nicely spotted. I have no clue why would it be on both dependencies...

PR #432, opened 2020-03-31
Remove rake development dependency

bumi:
We already have it as runtime dependency.

[closes: #426]


beckerei:
Is this supposed to be merged, and released?
Currently we have to explicitly downgrade rack in order to use latest version of sneakers, which comes with support for middlewares 🎉

michaelklishin:
Thank you!

PR #435, opened 2020-03-31
Update Ruby versions in travis.yml to run tests against latest ruby version on CI

bumi:
Running tests again the latest stable version 2.7 and the latest patch level version of previous ruby versions.

Available rubies on travis-ci can be found [here](http://rubies.travis-ci.org/).

Ruby 2.3.8 has reached EOL and does not get any updates anymore. This PR removes running tests against that version. Official support for not maintained ruby versions should be dropped.


rpbaltazar:
Shouldn't we update the Dockerfile image as well? It is pointing to alpine 2.3.7 still. 
I had [opened a PR](https://github.com/jondot/sneakers/pull/430/files) to update it to 2.4 but since you're making the changes here to travis, then i think it makes sense to send them all as one.

michaelklishin (approved):

michaelklishin:
Thank you. We will switch to Actions but bumping the Ruby versions we test against makes a lot of sense 👍

PR #441, opened 2020-06-15
Handle all exceptions

### LINKED ISSUE #412: Not StandardError exception during error reporting
Hello, I've found a bug that makes Sneakers not to report acknowledge state on error.

Wiki suggests to pass `context_hash` to error notification agents. [link](https://github.com/jondot/sneakers/wiki/Configuration#error-reporter)
```ruby
Sneakers.error_reporters << proc { |exception, _worker, context_hash| Honeybadger.notify(exception, context_hash) }
```
I'm using [Sentry raven](https://github.com/getsentry/raven-ruby) for error notifications. Sometimes it fails to serialize `context_hash` to json because `Bunny::Channel` instance in `delivery_info` of `context_hash` causes
```
SystemStackError: stack level too deep
```

As result:
1. RabbitMQ keeps message as Unacked
2. Such messages number may reach number of prefetched messages for worker
3. Worker hangs without performing any workload

Sneakers handles only `StandardError` and its descendants [here](https://github.com/jondot/sneakers/blob/master/lib/sneakers/error_reporter.rb#L19) and can not handle `SystemStackError` because it is not a `StandardError`.

Previously I [fixed similar bug](https://github.com/jondot/sneakers/pull/373) in another place. And I would be happy to do the same here.

But I think that it would be better to rescue from `Exception` in both places and re-raise exception in case of `Interrupt`, `NoMemoryError` and other exceptions that should not be swallowed.

WDYT?

sharshenov:
Previously Sneakers was able to [handle `StandardError` and `ScriptError`](https://github.com/jondot/sneakers/pull/373) exceptions and their descendants. But there are more exceptions that could cause problems.

* `SystemStackError` (stack level too deep)
* `NoMemoryError`
* custom descendants of `Exception` (e.g. declared by gems)

In case of these exceptions, the handler is unable to report job state to the RabbitMQ. RabbitMQ keeps sending new jobs to the Sneakers until prefetch value reached. It makes Sneakers worker hang without processing any new message.

Closes #412

### DIFF
--- a/lib/sneakers/worker.rb
+++ b/lib/sneakers/worker.rb
@@ -72,31 +72,34 @@ module Sneakers
           end
           res = block_to_call.call(deserialized_msg, delivery_info, metadata, handler)
         end
-      rescue StandardError, ScriptError => ex
+      rescue SignalException, SystemExit
+        # ServerEngine handles these exceptions, so they are not expected to be raised within the worker.
+        # Nevertheless, they are listed here to ensure that they are not caught by the rescue block below.
+        raise
+      rescue Exception => ex
         res = :error
         error = ex
         worker_error(ex, log_msg: log_msg(msg), class: self.class.name,
                      message: msg, delivery_info: delivery_info, metadata: metadata)
-      end
-
-      if @should_ack
-
-        if res == :ack
-          # note to future-self. never acknowledge multiple (multiple=true) messages under threads.
-          handler.acknowledge(delivery_info, metadata, msg)
-        elsif res == :error
-          handler.error(delivery_info, metadata, msg, error)
-        elsif res == :reject
-          handler.reject(delivery_info, metadata, msg)
-        elsif res == :requeue
-          handler.reject(delivery_info, metadata, msg, true)
-        else
-          handler.noop(delivery_info, metadata, msg)
+      ensure
+        if @should_ack
+          if res == :ack
+            # note to future-self. never acknowledge multiple (multiple=true) messages under threads.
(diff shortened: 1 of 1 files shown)


sharshenov:
[The failed job of Travis](https://travis-ci.org/github/jondot/sneakers/jobs/698588572) seems to be irrelevant to the PR.

lib/sneakers/worker.rb
```diff
@@ -72,31 +72,31 @@ def process_work(delivery_info, metadata, msg, handler)
           end
           res = block_to_call.call(deserialized_msg, delivery_info, metadata, handler)
         end
-      rescue StandardError, ScriptError => ex
+      rescue SignalException, SystemExit
+        # ServerEngine handles these exceptions, so they are not expected to be raised within the worker.
+        # Nevertheless, they are listed here to ensure that they are not caught by the rescue block below.
+        raise
+      rescue Exception => ex
```
kalys, line 79: What about interruption signals? Will they work?

sharshenov: @kalys Look at the [previous `rescue` block](https://github.com/jondot/sneakers/pull/441/files#diff-89d631e42c0689fc74b63e84d8605cdcR75). `Interrupt` is a subclass of `SignalException`.

sharshenov:
@michaelklishin sorry for tagging you, but this PR resolves a bug when errors such as `SystemStackError` may cause Sneakers consumer outage. Any chance of this change to be reviewed?

michaelklishin:
Thank you for spotting this.

PR #444, opened 2020-07-24
Configure workers run by rake sneakers:run

l-schirrmeister:
Adds the possibility to configure the workers, run by rake sneakers:run through Sneakers.rake_worker_classes

It will use still the workers from ENV[WORKERS]
But you can delete that and instead set Sneakers.rake_worker_classes = [Class1, Class2, ...] or set something that responds to :call.
If nothing is set, still Sneakers::Classes are used.

We have the situation that we would like to set the classes run by the task directly from within our project through a setting and not use an ENV for it, since that must be known from outside of it.
On the other hand, we use a base class for our workers so that the default strategy with Sneakers::Classes is not working for us as well.

### DIFF
--- a/lib/sneakers.rb
+++ b/lib/sneakers.rb
@@ -50,6 +50,14 @@ module Sneakers
     logger.level = loglevel
   end
 
+  def rake_worker_classes=(worker_classes)
+    @rake_worker_classes = worker_classes
+  end
+
+  def rake_worker_classes
+    @rake_worker_classes
+  end
+
   def logger=(logger)
     @logger = logger
   end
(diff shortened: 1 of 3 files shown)


sharshenov:
Hey @l-schirrmeister,

Have you tried passing a proc to a `Sneakers::Runner`?
```ruby
require 'sneakers/runner'

workers_proc = -> { [Foo, Bar, Baz] }

r = Sneakers::Runner.new(workers_proc)
r.run
```

[Details](https://github.com/jondot/sneakers/pull/210/files)

PR #446, opened 2020-08-08
maxretry configure_queue fix and specs

### LINKED ISSUE #442: MaxRetryHandler tries to set wrong x-dead-letter-exchange value
Looks like this was introduced in version 2.12.0.

I had previously defined retry queues and exchanges as follows:
exchanges:
  - activejob-retry
  - activejob-retry-requeue
  - activejob-error

queues:
  - activejob-retry
  - activejob-error

And from all other queues (lets say these are worker queues), the x-dead-letter-exchange is set to 'activejob-retry'

Now, the MaxRetryHandler has a new convenience method 'configure_queue' that tries to set x-dead-letter-exchange and it is using the queue names instead even though I have 'retry_exchange' configuration option defined.

Thus I get the following error: (lets say the queue is 'mailer')

Unexpected error PRECONDITION_FAILED - inequivalent arg 'x-dead-letter-exchange' for queue 'mailer' in vhost '/': received 'mailer-retry' but current is 'activejob-retry'

I am also passing the x-dead-letter-exchange value as arguments but it is ignored.

```
class MailerWorker
  include Sneakers::Worker
  from_queue :mailer, { handler: Sneakers::Handlers::Maxretry, retry_routing_key: 'mailer', arguments: { :'x-dead-letter-exchange' => "activejob-retry" } }
end
```
#376 introduced it
#427 does not fix it

dwkoogt:
maxretry configure_queue method fixed for following:
- retrieve retry-exchange from correct place on opts
- preserve other argument values
- do not overwrite user provided x-dead-letter-exchange value

resolves #442

### DIFF
--- a/lib/sneakers/handlers/maxretry.rb
+++ b/lib/sneakers/handlers/maxretry.rb
@@ -84,7 +84,11 @@ module Sneakers
 
       def self.configure_queue(name, opts)
         retry_name = opts.fetch(:retry_exchange, "#{name}-retry")
-        opts.merge(arguments: { "x-dead-letter-exchange": retry_name })
+        queue_opts = { arguments: {} }.merge(opts[:queue_options])
+        if queue_opts[:arguments][:'x-dead-letter-exchange'].nil? && queue_opts[:arguments]['x-dead-letter-exchange'].nil?
+          queue_opts[:arguments][:'x-dead-letter-exchange'] = retry_name
+        end
+        queue_opts
       end
 
       def acknowledge(hdr, props, msg)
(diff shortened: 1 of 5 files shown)


dwkoogt:
@michaelklishin ruby 2.3 test breaks due to pry-byebug. Downgrading pry-byebug to 3.2.0 will fix it. How do you want to proceed?

michaelklishin (changes requested):
Please undo the dependency change. Ruby 2.3 has been unmaintained since 2019. In general, sneaking in such changes into "bug fix PRs" is a sure way to have them rejected.

PR #455, opened 2021-06-21
 Handle all exceptions except SignalException and SystemExit 

aberranthacker:
Previously Sneakers was able to handle StandardError and ScriptError exceptions and their descendants. But there are more exceptions that could cause problems.

* SystemStackError (stack level too deep)
* NoMemoryError
* custom descendants of Exception (e.g. declared by gems)

In case of these exceptions, the handler is unable to report job state to the RabbitMQ. RabbitMQ keeps sending new jobs to the Sneakers until prefetch value reached. It makes Sneakers worker hang without processing any new message.


DarkSideF:
@jondot Hello! Can you please tell me what we should do for merge this PR?

PR #459, opened 2021-10-23
Github Actions

pedrofurtado:
@jondot 🤝 🍻 

Example of CI output: https://github.com/pedrofurtado/sneakers/actions/runs/1376170508


michaelklishin:
Thank you!

PR #466, opened 2022-11-08
fix: Unblock rake from v12 after forking as we are on v13 in some ser…

arielscherman:
…vices


simi:
@michaelklishin would it be welcomed to open PR to unlock rake? This prevents me from using latest sneakers in my project.

Issue #468, opened 2023-02-12
Ruby 3.2.0 File#exists?


wwwermishel:
Methods 'exists?'  for class File was removed in ruby 3.0.0 => https://bugs.ruby-lang.org/issues/17391
Could you please change method 'exists?' on 'exist?' here => https://github.com/jondot/sneakers/blob/31d0cb25dc5bbcfb0749567e9e0f80e6353fb66b/lib/sneakers/spawner.rb#L9
In Ruby 2.*.* we also hav method 'esixt?' for File.

michaelklishin:
You are welcome to submit a PR.

PR #469, opened 2023-02-12
Using File#exist? method instead of File#exists? / In ruby 3 File#exists? was removed

wwwermishel:
https://github.com/jondot/sneakers/issues/468


michaelklishin:
This PR and similar 3.0+ oriented ones make us consider if `main` can support older Rubies or should focus on 3.0+ exclusively.

Sadly Debian Bullseye and Ubuntu 20.04 still have 2.7.x releases.

